PR #4, opened 2017-11-01
8.4 Update (DO NOT MERGE)

kalpaitch:
image


kalpaitch:
8.4 runs with drush 8.1.15

There are a few entity updates that need to be done but they're not critical so will aim to get this upgrade in first.

PR #11, opened 2017-11-02
Add a CODEOWNERS file

chris-moreton:
I've added this primarily to make sure that I stay aware of any changes to DevOps stuff.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,21 @@
+# This file specifies which users will be automatically asked to review certain files
+# when a pull request is made that touches those files.
+
+############
+# Default
+############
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,23 @@
+# This file specifies which users will be automatically asked to review certain files
+# when a pull request is made that touches those files.
+
+############
+# Default
+############
+*                @kalpaitch @stemount @jon-edery @chris-moreton
```
kalpaitch: You probably don't want this much crossover with regards to the web application, you'll get spammed with requests @chris-moreton

chris-moreton: Yes, good point

PR #15, opened 2017-11-03
PAR-767: Updated role assignment logic.

kalpaitch:
Updated logic, but we also need to discuss the invite form as it's become a bit overcrowded.


kalpaitch:
Still testing this ticket and hoping to find a way to make invitations more testable.

PR #16, opened 2017-11-04
PAR-867 Enforcement officer details added to enforcement flows

jon-edery:
This Pr resolves cache override issues when multiple enforcement notifications are raised concurrently this was overriding the temp cache store and load incorrect values from other forms.

The correct Legal entity of the enforced organisation is now displayed throughout forms of the enforcement flow.

Displaying of the enforcement officer details throughout the flows has been added.

screen shot 2017-11-04 at 15 49 58

screen shot 2017-11-04 at 15 49 34

screen shot 2017-11-04 at 15 49 15


web/modules/features/par_enforcement_flows/src/Form/ParEnforcementRaiseNoticeDetailsForm.php
```diff
@@ -33,65 +35,59 @@ public function getFormId() {
...
+      '#markup' => $this->t('Include the following information'),
     ];
 
-    $form['organisation'] =[
+    $form['enforcement_text'] =[
       '#type' => 'fieldset',
       '#attributes' => ['class' => 'form-group'],
       '#collapsible' => FALSE,
       '#collapsed' => FALSE,
+      '#prefix' => '<ul>',
```
stemount: is this supposed to be an unordered list?

web/modules/features/par_enforcement_flows/src/Form/ParEnforcementRaiseNoticeDetailsForm.php
```diff
@@ -33,65 +35,59 @@ public function getFormId() {
...
+    $form['enforcement_title'] = [
       '#type' => 'markup',
-      '#markup' => $par_data_authority->get('authority_name')->getString(),
-      '#prefix' => '<div><h1>',
-      '#suffix' => '</h1></div>',
+      '#markup' => $this->t('Include the following information'),
     ];
 
-    $form['organisation'] =[
+    $form['enforcement_text'] =[
```
stemount: does 

`$form['enforcement_text'] = [
  '#theme' => 'item_list',
  '#items' => $enforcement_data,
];`

work?

jon-edery: yes however we need to do some front end changes as the item-list template adds a list class here -> {%- set attributes = attributes.addClass('list') %} prevents the li items being indented as per wire-frames.

web/modules/features/par_enforcement_flows/src/Form/ParEnforcementEnforceOrganisationForm.php
```diff
@@ -34,15 +34,16 @@ public function getFormId() {
    * @param \Drupal\par_data\Entity\ParDataPartnership $par_data_partnership
    *   The Partnership being retrieved.
    */
-  public function retrieveEditableValues() {
-
+  public function retrieveEditableValues(ParDataPartnership $par_data_partnership) {
+    $this->setState("edit:{$par_data_partnership->id()}");
```
kalpaitch: Only if there is a partnership id though.

web/modules/features/par_enforcement_flows/src/Form/ParEnforcementOfficerDetailsForm.php
```diff
@@ -0,0 +1,153 @@
...
+   */
+  public function retrieveEditableValues(ParDataPartnership $par_data_partnership) {
+
+    $this->setState("edit:{$par_data_partnership->id()}");
+
+    if ($enforcement_officer = $this->getEnforcingPerson()) {
+      // Load person data.
+      $this->loadDataValue("first_name", $enforcement_officer->get('first_name')->getString());
+      $this->loadDataValue("last_name", $enforcement_officer->get('last_name')->getString());
+      $this->loadDataValue("work_phone", $enforcement_officer->get('work_phone')->getString());
```
kalpaitch: We should have `$this->loadDataValue("enforcing_officer_id", $enforcement_officer->id());` and then we can just use `$this->getEnforcingOfficerId()` (like how you get the authority and organisation ids) instead of `$this->getEnforcingPerson()` after this point.

Basically just resolves any issues where someone else comes and removes the enforcing officer from the authority before they submit the application.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAuthoritySuggestionForm.php
```diff
@@ -38,14 +38,15 @@ public function titleCallback() {
    *   The Partnership being retrieved.
    */
   public function retrieveEditableValues(ParDataPartnership $par_data_partnership = NULL) {
-
+    $this->setState("edit:{$par_data_partnership->id()}");
```
kalpaitch: Yea you're right, this may be an issue.

1) we need to check to make sure there is a partnership id before setting the state like this.
2) We may need to systematically go through all forms and update the state values.

web/modules/features/par_enforcement_flows/src/Form/ParEnforcementEnforceOrganisationForm.php
jon-edery: the Partnership object is auto loaded via the route so there needs there has to be a partnership id passed through the URL here

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAuthoritySuggestionForm.php
jon-edery: All form that use this cache store have been updated.  A partnership object is enforced here via the param if an entity of type ParDataPartnership is passed through it will always have an id i.e. ->id() will always return a value for the state.

jon-edery: updated.

web/modules/features/par_enforcement_flows/src/Form/ParEnforcementOfficerDetailsForm.php
jon-edery: Done.

PR #20, opened 2017-11-07
Accessibility: Update titles across journeys (DO NOT MERGE YET)

stemount:
Provides a new section title in the flows.

PR #24, opened 2017-11-08
Accessibility: PAR-703 update application journey (DO NOT MERGE YET)

stemount:
This contains some changes to text in the application journey.

*Please consider/merge this*
https://github.com/UKGovernmentBEIS/beis-primary-authority-register/pull/20

### DIFF
--- a/web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsContactForm.php
+++ b/web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsContactForm.php
@@ -42,6 +42,21 @@ class ParPartnershipFlowsContactForm extends ParBaseForm {
     return 'par_partnership_contact';
   }
 
+  public function titleCallback() {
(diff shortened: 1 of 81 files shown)


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsContactForm.php
```diff
@@ -42,6 +42,21 @@ public function getFormId() {
     return 'par_partnership_contact';
   }
 
+  public function titleCallback() {
+    // Check if editing an existing entity.
+    $par_data_person = $this->getRouteParam('par_data_person');
+
```
jon-edery: I would add this logic in an else so we don't always run it then override if its not needed

stemount: This is a fair shout @jon-edery however the line below does give a default of "Add a contact"

PR #31, opened 2017-11-09
PAR-885 accessibility updates

jon-edery:


web/modules/features/par_enforcement_flows/src/Controller/ParEnforcementFlowsPartnershipPageController.php
```diff
@@ -49,15 +49,26 @@ public function content(ParDataPartnership $par_data_partnership = NULL) {
...
+      '#collapsible' => FALSE,
+      '#collapsed' => FALSE,
     ];
+
     $build['partnership_actions']['link'] = [
-      '#type' => 'markup',
-      '#markup' => t('@link', [
-        '@link' => $this->getFlow()->getNextLink('raise')->setText('Send notification of enforcement action')->toString(),
-      ]),
+      '#type' => 'fieldset',
```
kalpaitch: What's the reason for having a fieldset within a fieldset if they only have one element inside??

jon-edery: yea we should not this is an old element that got left in I have removed it.

PR #36, opened 2017-11-13
Entities now have their label in the edit link, leaving the field based things intact.

stemount:
See screenshot

messages image 1336026748


web/modules/custom/par_flows/src/ParDisplayTrait.php
```diff
@@ -164,6 +164,9 @@ public function displayEntityOperationLinks($section, $entity, $field, $delta =
       }
       if (in_array('edit-entity', $operations)) {
         $params[$entity->getEntityTypeId()] = $entity->id();
+
+        // Edit link to include the entity title.
+        $link_name_suffix = $entity->label();
```
kalpaitch: Why are we overwriting this? It's above on line 151.

PR #54, opened 2017-11-17
Restore order with the build

stemount:


.travis.yml
```diff
@@ -23,12 +23,19 @@ before_install:
...
+# restart servers after changing uid/gids.
+  - docker-compose stop
+  - docker-compose up -d
   - cd ..
 
 #-----------------------------------------------------------------------------
 # Install dependencies
 #-----------------------------------------------------------------------------
 
+install:
```
stemount: I'm not sure if this is correct, is this something you put in @chris-moreton ?

stemount:
ok this PR had a successful build lets get this in?

.travis.yml
chris-moreton: @stemount Yes, I split up the phases so that I could control them more easily in the debug session

.travis.yml
```diff
@@ -40,8 +47,13 @@ before_install:
 #-----------------------------------------------------------------------------
 
   - cp web/sites/example.settings.local.php web/sites/default/settings.local.php
-  - sudo bash ./tools/fix-permissions.sh --drupal_path=./web --private_files=./private
-  
+  - chmod -R 777 private
```
kalpaitch: @stemount Why have we taken these out of the fix-permissions script? Can we put them back in? Want to keep this all tidy.

.travis.yml
```diff
@@ -103,16 +114,14 @@ script:
 #-----------------------------------------------------------------------------
 # Generate HTML reports with screenshots. HTML report location: tests/reports/allure-reports/index.html, TAR it up.
 #-----------------------------------------------------------------------------
-  - travis_retry docker exec -ti par_beta_web bash -c "cd tests && if [ -d ./errorShots ]; then tar -cf reports/errorShots.tar.gz ./errorShots; fi"
-#  - travis_retry docker exec -ti par_beta_web bash -c "cd tests && ./node_modules/.bin/pa11y-ci -T 1 -j -s 'http://localhost/sitemap.xml' -f 'https://par-beta-test.cloudapps.digital' -r 'http://localhost:80' > ./reports/wcag2aa_report.json"
+#  - travis_retry docker exec -ti par_beta_web bash -c "cd tests && if [ -d ./errorShots ]; then tar -cf reports/errorShots.tar.gz ./errorShots; fi"
```
kalpaitch: Why is this commented out?

.travis.yml
```diff
@@ -145,14 +154,14 @@ deploy:
...
 
-addons:
-  artifacts:
-    debug: false
-    paths:
-      - ./tests/reports
-    target_paths: tests/$TRAVIS_BUILD_NUMBER
-    bucket: transform-par-beta-artifacts
-    s3_region: eu-west-1
+#addons:
```
kalpaitch: Why is this commented out?

web/modules/custom/par_actions/par_actions.module
```diff
@@ -55,11 +55,11 @@ function par_actions_tokens($type, $tokens, array $data, array $options, Bubblea
...
 
-function par_actions_cron() {
-  $scheduler = \Drupal::service('plugin.manager.par_scheduler');
-
-  foreach ($scheduler->getDefinitions(TRUE) as $definition) {
-    $plugin = $scheduler->createInstance($definition['id'], $definition);
-    $plugin->buildQueue();
-  }
-}
+//function par_actions_cron() {
```
kalpaitch: The solution would have been to disable the scheduler rule rather than to disable this function.

web/sites/default/settings.local.development.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 ini_set('memory_limit','512M');
-ini_set('max_execution_time', 90);
+ini_set('max_execution_time', 300);
```
kalpaitch: IMO I think if it takes longer than 60 seconds it should fail, I don't think it's helpful to have two different conditions for dev and prod like this.

.travis.yml
stemount: 777 is not really needed, nor the permissions script, hm

.travis.yml
stemount: @jaffamonkey i was pretty sure they were not producing reports and screenshots anymore

web/sites/default/settings.local.development.php
stemount: Yes, it was mostly the Travis thing causing an exit code 0.

web/modules/custom/par_actions/par_actions.module
stemount: Yes.

PR #57, opened 2017-11-19
Shaving PHPUnit build time from 6min down to 3min

stemount:


stemount:
@kalpaitch I am 10000 % with you, maybe it needs to be outside of script: and rather in an install: thing?! as its a dependency…

This was always set up in the wrong place in the script and here is a perfect opportunity to fix this

PR #62, opened 2017-11-20
PAR-913

jon-edery:
screen shot 2017-11-20 at 16 05 09

### DIFF
--- a/web/modules/custom/par_flows/src/ParDisplayTrait.php
+++ b/web/modules/custom/par_flows/src/ParDisplayTrait.php
@@ -374,20 +374,15 @@ trait ParDisplayTrait {
     foreach ($fields as $field_name => $view_mode) {
 
       $rendered_field = $entity->get($field_name)->view($view_mode);
-      $rendered_element =  $this->getRenderer()->render($rendered_field);
-
       $element[$field_name] = [
         '#type' => 'fieldset',
         '#collapsible' => FALSE,
         '#collapsed' => FALSE,
       ];
 
-      $element[$field_name]['value'] = [
-        '#type' => 'markup',
-        '#markup' => $rendered_element ? $rendered_element : '<p>(none)</p>',
-        '#prefix' => '<div>',
-        '#suffix' => '</div>',
-      ];
+      $element[$field_name]['value'] = $this->renderMarkupField($rendered_field);
+      $element[$field_name]['value']['#prefix'] = '<div>';
+      $element[$field_name]['value']['#suffix'] = '</div>';
     }
     return $element;
   }
--- a/web/modules/custom/par_flows/src/ParDisplayTrait.php
+++ b/web/modules/custom/par_flows/src/ParDisplayTrait.php
@@ -189,8 +189,30 @@ trait ParDisplayTrait {
    *
    * Display all of the fields on the given legal entity
    * with the relevant operational links.
+   *
+   * @param string $section
+   *   The section title to use for this field-set.
+   *
+   * @param Entity $entity
+   *   The entity containing the fields to be rendered.
+   *
+   * @param array $fields
+   *   An array containing the field names and display modes to be rendered.
+   *
+   * @param array $operations
+   *   An array required operations.
+   *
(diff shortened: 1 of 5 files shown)


web/modules/features/par_enforcement_flows/src/Form/ParEnforcementSubmitNoticeForm.php
```diff
@@ -83,7 +83,7 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
       $form[$enforcement_action->id()]['action_title'] = $this->renderSection('Proposed enforcement action', $enforcement_action, ['title' => 'title'], [], TRUE, TRUE);
       $form[$enforcement_action->id()]['action_regulatory_function'] = $this->renderSection('Regulatory function', $enforcement_action, ['field_regulatory_function' => 'summary'], [], TRUE, TRUE);
       $form[$enforcement_action->id()]['action_details'] = $this->renderSection('Details', $enforcement_action, ['details' => 'summary'], [], TRUE, TRUE);
-      $form[$enforcement_action->id()]['action_attach'] = $this->renderMarkupField($enforcement_action->get('document')->view('full'));
+      $form[$enforcement_action->id()]['action_attach'] = $this->renderedElementRenderSection('Attachments',$enforcement_action, ['document' => 'full'], TRUE);
```
stemount: would `$this->renderSection('Attachments', $enforcement_action, ['document' => 'full'], [], TRUE);` work here?

web/modules/custom/par_flows/src/ParDisplayTrait.php
```diff
@@ -328,4 +328,67 @@ public function renderPercentageTick($percentage = 0) {
...
+      $rendered_field = $entity->get($field_name)->view($view_mode);
+      $rendered_element =  $this->getRenderer()->render($rendered_field);
+
+      $element[$field_name] = [
+        '#type' => 'fieldset',
+        '#collapsible' => FALSE,
+        '#collapsed' => FALSE,
+      ];
+
+      $element[$field_name]['value'] = [
```
stemount: failing my comment at the end of the file being nonsense, check out `$this->renderMarkupField`

web/modules/features/par_enforcement_flows/src/Form/ParEnforcementSubmitNoticeForm.php
jon-edery: no it does not render correctly.

stemount: :+1:

kalpaitch: I'd quite like to be able to see why it does not render correctly, because correcting it if poss is probably better than creating a new method.

I'll be able to test this afternoon.

kalpaitch: One of the reasons we have a centralized method for this is so that it adds stuff like pagination, I don't see this handled here atm...

kalpaitch:
Also can you discuss with @jaffamonkey a way to test for this. If it's different the we should make sure it doesn't regress.

web/modules/custom/par_flows/src/ParDisplayTrait.php
```diff
@@ -189,8 +189,30 @@ public function displayEntityOperationLinks($section, $entity, $field, $delta =
...
+   *
+   * @param Boolean $single
+   *
+   * @param Boolean $section_title_only
+   *    An option to only output a title with the correct markup.
+   *
+   * @return mixed
    */
-  public function renderSection($section, $entity, $fields, $operations = [], $title = TRUE, $single = FALSE) {
+  public function renderSection($section, $entity, $fields, $operations = [], $title = TRUE, $single = FALSE, $section_title_only = FALSE) {
```
kalpaitch: Do we need this extra parameter now??

PR #63, opened 2017-11-21
Trading Name form contextual titles and presentation tweaks

stemount:
- Improving presentation slightly by spacing out form elements.

messages image 2929832291


kalpaitch:
Shouldn't we be doing this consistently, doing this in one place and not others may look a bit odd??

stemount:
@kalpaitch yeah, other forms have got add/edit titles already, just in the interest of small PRs…

kalpaitch:
Oh I see, it's the spacing at the bottom. I guess if you want to get this in quick then sure, but we kinda really want to tidy up the display of our forms. I'm going to have a bit of a think about this too and put it in the technical debt epic.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsTradingForm.php
```diff
@@ -14,15 +14,26 @@ class ParPartnershipFlowsTradingForm extends ParBaseForm {
...
     return 'par_partnership_trading_name';
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function titleCallback() {
+    $trading_name_delta = $this->getRouteParam('trading_name_delta');
+
+    $action = $this->t(is_numeric($trading_name_delta) ? 'Edit' : 'Add another');
```
kalpaitch: I think it's best to judge the action/title based on the route and flow. The section title needs to be based on flow, and the page title needs to be based on route. This the best way to achieve this for now is to have `if routename === par_partnership_flows.trading_name_add then 'add' else' edit'`. We can integrate a better way soon like have titles on the flow steps themselves.

And it's difficult to tell whether it's add another or just add. Might be better to just call it add if that's ok with the design.

stemount: I think its pretty reliable checking if there is an entity loaded in the context of the page to do with that form, but I hear you about the 'add another' thing, maybe lets just make this more generic? Add…

It's definitely possible though to load the partnership in this case and figure out if it already has a trading name should we want this? @kalpaitch @nao360 ?

nao360: @stemount we can use 'Add a' instead of 'Add another' - agreed it works in all scenarios, particularly when the user is adding the very first one. I will update this change in the UX mockups.

jon-edery: Agree if we can get away with providing a more generic title here it will avoid the issue.

stemount:
@kalpaitch lets keep it uniform on how we have done add/edit throughout, we lean on the Drupal requirements system for checking its a number etc and the form can still be reused without adding any other route names.

I do agree it could possibly look for `add` but in the absence of a trading name field number it is add.

kalpaitch:
This PR is super polluted now. Can we get it back to just the stuff for this ticket.

PR #65, opened 2017-11-21
PAR-954: Updated scheduler manager and improved error handling.

kalpaitch:
* Updated schedule manager with a minimum interval so that it does not run more than once an hour, this allows any actions that it added to a queue to be processed before it is next run.
* Separated actions from the queue so that the schedule manager can run actions directly rather than add them to the queue, this is useful for enforcement notices as we'll rarely be doing that many at a time.
* Created one generic queue for processing all actions (added action plugin id to the data) so that we don't need to create new queue integrations when we create new actions.
* Changed auto-approval time as per PAR-976 to five inclusive days and re-enabled the auto-approver of ENs.

To run go to `/admin/config/system/cron/jobs` and run the 'Process all scheduled PAR rules' job. You will need to have ENs with actions that are awaiting approval.


web/modules/custom/par_actions/par_actions.module
```diff
@@ -55,11 +53,10 @@ function par_actions_tokens($type, $tokens, array $data, array $options, Bubblea
...
 
-//function par_actions_cron() {
-//  $scheduler = \Drupal::service('plugin.manager.par_scheduler');
-//
-//  foreach ($scheduler->getDefinitions(TRUE) as $definition) {
-//    $plugin = $scheduler->createInstance($definition['id'], $definition);
-//    $plugin->buildQueue();
-//  }
-//}
+function par_actions_cron() {
```
jon-edery: function comments...

web/modules/custom/par_actions/src/Annotation/ParSchedulerRule.php
```diff
@@ -79,6 +79,14 @@ class ParSchedulerRule extends Plugin {
    */
   public $time;
 
+  /**
+   * Whether to resolve this action by sticking it in a queue
+   * instead of trying to resolve immediately.
+   *
+   * @var boolean
+   */
+  public $queue;
```
jon-edery: $should_queue might be less confusing

web/modules/custom/par_actions/src/ParScheduleManager.php
```diff
@@ -76,4 +96,29 @@ public function getDefinitions($only_active = FALSE) {
     return $definitions;
   }
 
+  /**
+   * A helper method to run any plugin instance.
+   *
+   * @param $definition
```
jon-edery: what is definition?

web/modules/custom/par_actions/src/Annotation/ParSchedulerRule.php
kalpaitch: I'll come back to this, I'm thinking 'queue_process' might be more applicable, but not sure yet.

web/modules/custom/par_actions/src/ParScheduleManager.php
kalpaitch: Added description.

PR #67, opened 2017-11-22
The 'summary of enforcement notice' appears to be truncated as it was…

stemount:
… using the trimmed text filter


kalpaitch:
Let's add some tests for this to ensure we are testing all text area fields with blocks of text longer than 1000 characters.

PR #73, opened 2017-11-24
Reinstate reports from wdio/phpunit and wdio error screenshots.

stemount:
- also adds new variables for ClamAV auth for testing filefield
- paratest produces report and used to exit with a 127, but now correctly exits as 0
- added json report to wdio.config to ensure report is exported


.travis.yml
```diff
@@ -159,11 +163,14 @@ addons:
...
+env:
+  global:
+  - secure: Fs3vok73zakpxIio13QqNuGlUoLVGeaE85FbFTTQI3+xdHWzVGausI7HW7GRh32cQP3jgPazcIe4WE/og4FHWwPIJzRzknbkaBN6zsmETS2nhh7H8rNCbw/vXqC1hXhXK3VdaIvuL9t+UBaqeojrgIvF7DVs5G3shtKX6om5NscxYDUmIMPeQL5gaIXVcGGIVlyL7rruQQ8swt1ld9YtNhoGv9zhJjezr5r2qGwrJkx2JkPWJJre6Dy8PaPWaOKQ9vV3NtI+uG2mbctBkBqG05XZ1Rm3V8r8lOyDhpQcWkn8M616QMkd/bWZgcS97RX7MEdKBQIhEGFAo66z76Uc+PwgnTy8hF3pM6oF1adrw9GBfbMD5gsS/KVLjs/GOlHF3blsoFALk99K/6WIt3/5682LmXxvY56lnJOBMWgDOiWk8dlmj5CXLnxEN8GCgMDphV+xkv8luN3H18dtnHQ9dp8p1w4CNwopsUXW/PgOdQput8DRP74Y0D3cpGMjoHEhY1kxuWghwy2q+SHCgrvMzjyZs4EXWJSQiGQmqw5AjUzntR2BsqEsne78L5SBC6rCcMalTcWf+BSHITI75RiEdzUg+wSciV5FMWwD1OBlAqgjeOpocWcoTvUYQW4jl6HnwWyOuIKYPTZnkqsKQjDsrfLsAgxtVZNRh4E6/y3ZJBY=
```
kalpaitch: What is this btw?

.travis.yml
```diff
@@ -140,16 +144,16 @@ after_script:
...
-    secret_access_key: $S3_SECRET_KEY
-    bucket: "transform-par-beta-artifacts"
-    upload-dir: builds
-    skip_cleanup: true
-    local_dir: $TRAVIS_TAG
-    on:
-      repo: UKGovernmentBEIS/beis-primary-authority-register
-      tags: true
+- provider: s3
+  access_key_id: "$S3_ACCESS_KEY"
```
kalpaitch: Is this really needed, it used to work unquoted before.

.travis.yml
```diff
@@ -26,6 +26,8 @@ before_install:
 # set web server / docker root uid/gid to match Travis uid.
   - docker exec -t par_beta_web bash -c "usermod --uid=2000 www-data"
   - docker exec -t par_beta_web bash -c "groupmod --gid=2000 root"
+  - docker exec -t par_beta_web bash -c "adduser root www-data"
```
kalpaitch: @chris-moreton happy with this?

.travis.yml
stemount: it's actually a CLAMAV http_user and http_password thing. I notice we basically never had this in the repository settings. You can use `travis encrypt` to put these in the Travis file so you track what you have.

.travis.yml
stemount: sure, but its what the Travis tool reformatted it as.

stemount:
@jaffamonkey hi - i've fiddled with wdio here so wanted to get your approval on this. the errorShots and the report.json now generate things.

I would suggest the allure report is not suitable for us because it generates a HTML file which actually uses file:/// URIs which doesn't even work in the AJAX interface unless you serve it with a web serer…

.travis.yml
kalpaitch: @chris-moreton Just wondering if these should live in vault?

kalpaitch:
Yup, think it'd be good for @jaffamonkey and @chris-moreton to review before merging.

.travis.yml
chris-moreton: @kalpaitch Yep they are in the Vault already

chris-moreton: @kalpaitch The practice we are currently using is that the variables needed for Travis are in the Vault and then we add these to the Travis settings rather than store them in the yml file.

kalpaitch: @stemount do you mind updating these vars...

PR #75, opened 2017-11-24
Permissions fix in deploy script

chris-moreton:
These are two changes to the push.sh script that fix permissions for pushing to PaaS.


kalpaitch:
ATM this doesn't do anything @chris-moreton ??

kalpaitch:
Already in.

PR #76, opened 2017-11-27
PAR-912: Blocked

jon-edery:
screen shot 2017-11-27 at 10 27 24


kalpaitch:
Oh also there were quite a lot of notices/warnings when generating this page now..??
screenshot from 2017-11-27 11-00-13

jon-edery:
As per comments there is some tech debt around the current way we our rendering markup and the need for some class to deal with some edge cases.   This ticket is blocked as a result.

PR #82, opened 2017-11-28
PAR-932 title callback additions

jon-edery:


web/modules/features/par_enforcement_flows/src/Form/ParEnforcementConfirmNoticeForm.php
```diff
@@ -36,6 +36,14 @@ public function accessCallback(ParDataEnforcementNotice $par_data_enforcement_no
     return parent::accessCallback();
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function titleCallback() {
+    $this->pageTitle = "Confirmation| Enforcement action decision";
```
stemount: this should be `return "Confirmation | Enforcement action decision";` as if this flow had a `default_section_title` in the yml file, it would possibly have `Blah default title | Confirmation | Enforcement action decision` as a title

kalpaitch:
@jon-edery The steps we've used in other places to test the title have looked like...
```

PR #87, opened 2017-12-01
[PAR-1013] Enforcement Notice send/receive views (DO NOT MERGE)

stemount:
**Summary:**
Create a new split view for enforcement notices.
/enforcement-notices/sent
/enforcement-notices/received

**Proposed technical tasks/solution:**
- [x] Create contextual filter with a default argument of the current logged in users PAR Authorities to filter by fields Primary Authority and Enforcing Authority.
- [x] Prevent memberships lookup function from following enforcement notices as it will add erroneous partnerships to the partnerships list.
- [x] Restrict view permissions to the correct roles/permissions.

**Testing notes:**
As an enforcement officer (with just EO role), log in to the system and confirm you can send an enforcement notice and it shows in your Sent EN view.

As a PA officer you should be able to see Sent messages from your Primary Authorities you are a member of as well as see received ENs to your primary authorities. You should be able to see ones that are sent from your PA.

**Changes:**
primary_authority_register___primary_authority_register

enforcement_notifications___sent___primary_authority_register

enforcement_notifications___received___primary_authority_register


kalpaitch:
WTF is going on with this ticket :) Can we clean it up so it's not 100+ commits and over 70 files??

PR #88, opened 2017-12-01
PAR-859: Created notifications for par data entity events (BLOCKED)

kalpaitch:
The par_notification module i split into the following parts:
1) The par_notifications.notifier service which is responsible for sending the notification (atm only has one e-mail plugin)
2) The message entities where we can store the messages that need to be sent.
3) the message manager which is responsible for loading these messages and tokenizing them as appropriate.
4) The ParDataEvent (new events) listeners which react on certain actions (currently only create, edit, delete) on a par_data entity and have to decide whether to send a notification, and to who. There should be one listener per action/message.

The main difficulty with this at the moment is testing because none of our test environments are configured in such a way that we can see the e-mails sent out. This is technical debt mentioned here - https://regulatorydelivery.atlassian.net/wiki/spaces/PA/pages/125108286/Technical+Debt#TechnicalDebt-Emailhandling


kalpaitch:
@jon-edery @stemount would you review please. Is pretty basic and works OK...

Apart from issues around data, basically getting the right tokens is tricky and might need to discuss further.

kalpaitch:
Phasing this approach in favour of the message stack modules - https://www.drupal.org/node/2180145

See new PR in the working - https://github.com/UKGovernmentBEIS/beis-primary-authority-register/compare/master...PAR-859.3

PR #89, opened 2017-12-01
PAR-947: Correcting issues sending emails.

kalpaitch:
This is related to two main issues:
1) Travis needs to know about notify.
2) There's a bug in the devel that crashes sending the invitation (see patch link below).
screenshot from 2017-12-01 16-45-50


kalpaitch:
See https://www.drupal.org/project/devel/issues/2927963

PR #93, opened 2017-12-04
PAR-994 removed trimmed fields from notification summary

jon-edery:


kalpaitch:
a) does this not need to be the other view modes?
b) will need to export to feature.

jon-edery:
We only use the full display mode for this field but yes we can update all display modes as we dont use trimmed version anywhere.

PR #96, opened 2017-12-05
[PAR-1025 and PAR-1030]: Temp fix to remove premises data.

kalpaitch:
This temporarily removes lookup on all premises data.


kalpaitch:
Looks good to me.

PR #98, opened 2017-12-06
PAR-1009: Updated cron frequencies and errors.

kalpaitch:
Patch here will also need review - https://www.drupal.org/project/drupal/issues/2928889#comment-12372748

This can be tested by going to http://par.localhost/admin/config/system/cron/jobs and running cron (each job separately if you have to). There should be no errors or red crosses by the end.


stemount:
Cool - stopping unused cron disabling is a good shout.

PR #104, opened 2017-12-09
New Vagrant box and a helper script for rebuilding the environment

chris-moreton:
To test this:

1. Checkout the branch
2. cd ..
3. Add a line to vagrant.sh to checkout the branch after it clones the repo
4. sh beis-primary-authority-register/vagrant.sh

The new Vagrant box differs only in that it has all the Docker packages pre-installed.

The new vagrant.sh file must be run from the directory above the repo. It will do a few helpful things

1) Save your environment file
2) Destroy the VM and clear away conflicting processes
3) Install Composer dependencies on the host and the fix the permissions for the VM and Docker

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,29 +1,19 @@
 $script = <<SCRIPT
-echo I am provisioning...
-date > /etc/vagrant_provision_start
-
-curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
-sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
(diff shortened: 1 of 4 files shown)


chris-moreton:
@nao360 Yes, sorry, the process is predicated on the directory being named the same as the repo. This is a hangover from the name changing when we switched to BEIS GitHub.

The line 29 works because of what happens earlier in the script when it copies the .env file up to the parent directory.

SSH keys not working is strange - it sounds like perhaps you're not correctly set up somehow - I'll figure it out.

Yes, don't approve yet. I will modify to take into account a different directory name.

nao360:
@chris-moreton confirming it works me too.

PR #107, opened 2017-12-11
PAR-1013: Readded tests without commenting out.

kalpaitch:
Excluding spike tests in #106


jon-edery:
looks good Pending test results

PR #110, opened 2017-12-14
PAR-1048 needs lock file update

stemount:
Small error in previous PR


stemount:
actually, this is not needed.

PR #111, opened 2017-12-14
PAR 1061: Docker Images + seed data + deployment tools in VM

chris-moreton:
A new VM is referenced in the Vagrantfile. This VM includes all the command line tools needed to perform a deployment. Deployments are performed using "./push.sh" within the "cf" directory.

The Docker images are also included, meaning that they don't need to be downloaded every build.

The database of the Postgres docker container is also there - this is the new Drupal install database, which previous we needed to import before being able to run the command to import the required database from S3.

The initial download of the VM will take a little time, but once it's down, it's down. Subsequent builds will be super quick in comparison to the previous VM.

To test, verify that the following command builds the VM correctly. This will use this branch to build the VM.

Move to the directory above the repo and then:

sh <par_repo_dirname>/vagrant.sh PAR-1061


README.md
```diff
@@ -1,5 +1,7 @@
 # Department of Business Energy and Industrial Strategy - Regulatory Authority
 
+Test
```
stemount: Test? :-D

.travis.yml
```diff
@@ -187,4 +195,4 @@ env:
...
+  - secure: G/sN3FBcz3aBZgZ/rkCHc+Xrv/oVU2VUFtLmYH6A/kKLtYAckgn5Lj9FYrW545TBzr0ZumEp2/N2446xCzG4fKmU7bkFHAsbWrWJXXCF1ZQnEG1uhy33OtW2ubR4jzbUJL1qyF5JvCpers59I7dCVxQLfeZYqt0Tex87w5TZJqZzVKL6nEvnoTtrHrbk8SsaMGhkbtf+OKezO5SemT0FH9Jh4sXzmyByIXudgvfBCHTbIGqO+oFyyEhWQP84pB/fvkuppA6k74P6iscY3CwwJMvSFoQa3m4edMlEyEsgUEhbmP5TFagNl+baTWN3XS/hKOSIOm1HttsOKyfgxUzKQ4IYb9vUOl1d2CIvMWDpAgJOs8plNfHWbmZYw5SVpxp25Y/VtsfcFEFF2zYzF93swpiokchjcLVmDAVagG6repmy/VuwzoxbgaVUbbAU+sndzmdOTYBb8i1VpoGUKTpiHQOcdkDf3MDHi7fJnvgsRu/Mz+xfQ4y0Tm8OejnCzAzfiIuaAo06a0osp6NIbevqW85Z4Ag76XxxFsmfT7x3qSAGiSc9CJ0KUtg3usAi2EbixFtBk3mQMsILPOmwyhEqUMTGSX0zsM6wf+J1I58NSzoiGMo7mkFEK4wg5+aw+DbF3mLCBLRRsW+pyFegGaq2itaIZtOXMDKFsE85z6qHIL0=
```
stemount: this file is just missing an EOF line break

stemount:
> The Docker images are also included, meaning that they don't need to be downloaded every build.

This sounds good, it does mean we will need to maintain the Vagrant box for now, but as this is what we are using for development team-wide, this is a good idea. (Saves 500MB and 2 minutes!)

> A new VM is referenced in the Vagrantfile. This VM includes all the command line tools needed to perform a deployment. Deployments are performed using "./push.sh" within the "cf" directory.

Great, its good that we build from this rather than from a local machine.

chris-moreton:
@stemount 

> Let's review the deleting repository and vagrant upping. This seems a bit heavy handed and maybe we can work out an alternative solution. 

Ok. Perhaps we could get some scenarios where a developer would want to rebuild in such a situation. The script was primarily designed to destroy everything and start from scratch in the case where everything had gone pear-shaped. Do we perhaps need an additional script that would address only non-fatal scenarios, perhaps just performing a dependency refresh?

> Deleting the repository will lose developers local branches and git stashes. Deleting the Vagrant VM will also remove all history from the bash console of frequently used commands.

Yes, this is very valid, although sometimes destruction is necessary if the problems being experienced are permissions-related or completely unknown. Maybe we can parameterise the script to indicate what needs to happen, e.g.

vagrant-rebuild.sh [options]

-v : Destroy VM
-c : Rebuild composer dependencies
-n : Rebuild node dependencies
-a : Destroy everything, including repository, re-clone and rebuild

stemount:
Closing due to inactivity for a while, lets discuss tomorrow if this is something to go forward with.

kalpaitch:
Reopening, can be closed or merged following discussion :)

PR #113, opened 2017-12-19
Revoke mega commit

stemount:


sync/par_data.par_data_partnership_type.partnership.yml
```diff
@@ -38,6 +38,7 @@ configuration:
       confirmed_authority: 'Confirmed by the Authority'
       confirmed_business: 'Confirmed by the Organisation'
       confirmed_rd: Active
+      revoked: Revoked
```
kalpaitch: Minor point. There are certain special statuses that are available on all entities, we shouldn't have to include them here.

web/modules/custom/par_data/src/Entity/ParDataEntity.php
```diff
@@ -23,6 +23,18 @@ class ParDataEntity extends Trance implements ParDataEntityInterface {
   const REVOKE_FIELD = 'revoked';
   const ARCHIVE_FIELD = 'archive';
 
+  /**
+   * {@inheritDoc}
+   */
+  public function save() {
```
kalpaitch: That's exactly the idea... but we'd be able to only set revisions when the status changed.

web/modules/custom/par_data/src/Entity/ParDataEntity.php
```diff
@@ -325,6 +354,53 @@ public function setParStatus($value) {
     }
   }
 
+  public function getParStatusHistory($sort = SORT_ASC) {
```
kalpaitch: The only think I wonder is if we can find/make a table where all this information is more easily accessible, atm we have to pull out all revisions before querying their field properties.

PR #117, opened 2017-12-29
PAR-1059 & PAR-1062: Added pages for partnership revocation.

kalpaitch:
This is the initial work to create the pages for the revocation journey (just the pages and the flow, no link to it yet so the journey can't really be used even if we merge this):

screenshot from 2017-12-29 12-10-20
screenshot from 2017-12-29 12-10-03

The titles and various texty bits on the pages are a bit of a mess really, there's very little consistency of wording between these two pages or even between the approve journey but we can address when everyone is in next week.

### DIFF
--- a/sync/par_flows.par_flow.revoke_partnership.yml
+++ b/sync/par_flows.par_flow.revoke_partnership.yml
@@ -1,0 +1,26 @@
+uuid: 289f26b0-bee0-4c2d-b12e-3cfe96c2546f
+langcode: en
+status: true
+dependencies: {  }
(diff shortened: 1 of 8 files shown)


stemount:
Adding the routing and Views additional links in another PR.

kalpaitch:
@stemount please explain. I know this isn't all the work you did, but this PR specifically addresses tickets PAR-1059 and PAR-1062.

stemount:
@kalpaitch this was a mistake, hit close, not merge!

PR #119, opened 2017-12-31
PAR-1078: Enable revisioning for PAR entities with a `status_field`

stemount:
Create revision, set revision message in any previous hooks, save timestamp and set revision author to current Account ID.

### DIFF
--- a/web/modules/custom/par_data/src/Entity/ParDataEntity.php
+++ b/web/modules/custom/par_data/src/Entity/ParDataEntity.php
@@ -40,6 +40,18 @@ class ParDataEntity extends Trance implements ParDataEntityInterface {
   }
 
   /**
+   * {@inheritDoc}
+   */
+  public function save() {
+    // Create a revision when a PAR entity is saved.
+    $this->setNewRevision(TRUE);
+    $this->setRevisionCreationTime(REQUEST_TIME);
+    $this->setRevisionAuthorId(\Drupal::currentUser()->id());
+
+    return parent::save();
+  }
+
+  /**
    * {@inheritdoc}
    */
   public function getViewBuilder() {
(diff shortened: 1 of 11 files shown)


web/modules/custom/par_data/src/Entity/ParDataEntity.php
```diff
@@ -39,6 +39,18 @@ public function getTypeEntity() {
     return $this->type->entity;
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  public function save() {
+    // Create a revision when a PAR entity is saved.
+    $this->setNewRevision(TRUE);
```
kalpaitch: This is configurable, doesn't need to be done like this.

kalpaitch:
@stemount all you need to do to create a revision by default is put the 'revision' key on the entity type configuration...

```
langcode: en
status: true
dependencies: {  }
id: partnership
label: Partnership
description: 'A Partnership is a relationship between a Primary Authority and either a Business ("direct partnership") or a Co-ordinator ("co-ordinated partnership").  Note that in the latter case, the Business records may or may not be held in the PAR3 database.'
help: ''
revision: true
isDeletable: true
isRevokable: true
isArchivable: false
configuration:
  entity:
    label_fields:
      - 'Partnership between'
      - field_authority
      - and
      - field_organisation

**Comment by stemount** [comment]
Good shout @kalpaitch - definitely agree if configurable no need to override the method, lets discuss which ones should be revisioned.

File: web/modules/features/par_data_config/config/install/par_data.par_data_deviation_request_type.deviation_request.yml [hunk 2]
```diff
@@ -5,6 +5,7 @@ id: deviation_request
 label: 'Deviation Request'
 description: ''
 help: ''
+required: true
```
kalpaitch: hehe I think you mean 'revision' not required...

stemount:
Schoolboy error indeed with `required` @kalpaitch, will update feature too so the tests pass.

stemount:
As @kalpaitch tested, this second solution doesn't work at the minute (for some daft reason). We can look in the commit history just in case.

PR #120, opened 2018-01-02
PAR-1060 & PAR-1043: Partnerships that are in progress cannot be revoked.

kalpaitch:
Added an `inProgress` method for determining when an entity may be revoked, archived or deleted.

However, because we're using access control to deny access to the page if the operation cannot be performed then the actual message in the UI (PAR-1060) never really gets displayed, what happens instead is that the link is never offered on the dashboard. This has been the practice for the approval journey and various other journeys so I think best to stick with this, but it does change the A/C a little.

### DIFF
--- a/web/modules/custom/par_data/src/Entity/ParDataEnforcementAction.php
+++ b/web/modules/custom/par_data/src/Entity/ParDataEnforcementAction.php
@@ -68,7 +68,6 @@ class ParDataEnforcementAction extends ParDataEntity {
   const APPROVED = 'approved';
   const BLOCKED = 'blocked';
   const REFERRED = 'referred';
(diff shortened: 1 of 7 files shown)


web/modules/custom/par_data/src/Entity/ParDataEnforcementAction.php
```diff
@@ -211,12 +210,24 @@ public function refer($refer_notes) {
...
+      return TRUE;
+    }
+
+    return parent::inProgress();
+  }
+
+  /**
+   *  Get the referred note data from the current action.
+   *
+   * @return String referred_text | NULL
```
stemount: really unimportant spacing

PR #123, opened 2018-01-02
PAR-1045: Added message for partnership revocation.

kalpaitch:


web/modules/custom/par_notification/src/EventSubscriber/PartnershipRevocationSubscriber.php
```diff
@@ -0,0 +1,143 @@
...
+
+  /**
+   * @param ParDataEventInterface $event
+   */
+  public function onPartnershipRevocation(ParDataEventInterface $event) {
+    /** @var ParDataEntityInterface $enforcement */
+    $partnership = $event->getData();
+
+    // Only act on partnership entities.
+    if (!$partnership || $partnership->getEntityTypeId() !== 'par_data_partnership') {
```
jon-edery: My understanding from this is that par_data.entity.update event is triggered for all par data entities not specifically for entities of type partnership in which case onPartnershipRevocation is triggered every time any system entity of type ParData is updated. is this correct?

Would it make more sense performance wise to dispatch a more  specific event that will trigger when partnership entities are updated/revoked?

web/modules/custom/par_notification/src/EventSubscriber/PartnershipRevocationSubscriber.php
```diff
@@ -0,0 +1,143 @@
...
+      return;
+    }
+
+    // Load the message template.
+    $template_storage = $this->getEntityTypeManager()->getStorage('message_template');
+    $message_template = $template_storage->load(self::MESSAGE_ID);
+
```
jon-edery: Will this code execute when a partnership entity is updated but not revoked is that possible currently?

web/modules/custom/par_notification/src/EventSubscriber/PartnershipRevocationSubscriber.php
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * The events to react to.
+   *
+   * @return mixed
+   */
+  static function getSubscribedEvents() {
+    // Revocation event should fire after most default events to make sure
+    // revocation has not been cancelled.
+    $events[ParDataEvent::UPDATE][] = ['onPartnershipRevocation', -100];
```
jon-edery: what is the -100 used for? out of interest

kalpaitch: Priority. Giving this a low priority so that most other events happen first. It's not particularly relevant though.

web/modules/custom/par_notification/src/EventSubscriber/PartnershipRevocationSubscriber.php
kalpaitch: The loading of the template will yes, but obviously nothing thereafter. Probably best to move it like you say. But I think there are bigger improvements to make as you've mentioned above.

web/modules/custom/par_notification/src/EventSubscriber/PartnershipRevocationSubscriber.php
kalpaitch: Been thinking about this a while, and for the sake of simplicity have kept it like this until we can work out what we actually need. I think there are possibly some better options than event listeners within the notification stack, this was just the simplest.

kalpaitch: The other reason I'm not sure about using our own events (and actually wanted to get rid of them eventually) is that there seems to be some progress on an Entity Event API - https://www.drupal.org/project/entity/issues/2658136

kalpaitch:
Further improvements to be spiked in [PAR-1079](https://regulatorydelivery.atlassian.net/browse/PAR-1079)

PR #126, opened 2018-01-04
PAR-1057 RD dashboard: Add 'Revoked' to partnership search filter and link to revoke partnership.

stemount:
cursor_and_rd_help_desk___dashboard___primary_authority_register

See screenshot for the revoked search filter option and the revoke partnership link.

### DIFF
--- a/sync/views.view.helpdesk_dashboard.yml
+++ b/sync/views.view.helpdesk_dashboard.yml
@@ -1139,7 +1139,7 @@ display:
           element_wrapper_type: ''
           element_wrapper_class: ''
           element_default_classes: true
-          empty: N/A
+          empty: ''
           hide_empty: false
           empty_zero: false
           hide_alter_empty: true
--- a/sync/views.view.helpdesk_dashboard.yml
+++ b/sync/views.view.helpdesk_dashboard.yml
@@ -107,6 +107,7 @@ display:
             partnership_type: partnership_type
             partnership_status: partnership_status
             par_flow_link: par_flow_link
+            par_flow_link_1: par_flow_link
           info:
             id:
               sortable: false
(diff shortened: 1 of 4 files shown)


stemount:
@nao360 is an empty "actions" column good? Let's discuss

sync/views.view.helpdesk_dashboard.yml
```diff
@@ -107,6 +107,7 @@ display:
             partnership_type: partnership_type
             partnership_status: partnership_status
             par_flow_link: par_flow_link
+            par_flow_link_1: par_flow_link
```
kalpaitch: Could we change these two field ids to something meaningful.

web/modules/features/par_rd_help_desk_flows/src/Form/ParRdHelpDeskApproveConfirmForm.php
```diff
@@ -35,6 +35,11 @@ public function titleCallback() {
    * {@inheritdoc}
    */
   public function accessCallback(ParDataPartnership $par_data_partnership = NULL) {
+    // If partnership has been revoked, we should not be able to approve it.
```
kalpaitch: This is not part of this ticket, a) let's separate and/or b) let's put a massive @todo to say that this is a temporary fix and that properly controlling whether a transition from one state to another is more involved (stopping it here won't stop it happening).

stemount: Yes, it is a seperate ticket probably (and certainly nothing to do with the views presentation) but it is loosely related because revoke in general has pretty much broken this code @kalpaitch

sync/views.view.helpdesk_dashboard.yml
stemount: Are you sure you want to change views auto generated code? The next features export will probably override any changes.

PR #129, opened 2018-01-05
PAR-1078: Revisioning on status change.

kalpaitch:


web/modules/custom/par_data/src/Entity/ParDataEntity.php
```diff
@@ -462,6 +476,16 @@ public function getCompletionPercentage($include_deltas = FALSE) {
...
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function setNewRevision($value = TRUE) {
+    $this->setRevisionCreationTime(REQUEST_TIME);
+    $this->setRevisionAuthorId(\Drupal::currentUser()->id());
+
+    parent::setNewRevision($value);
```
stemount: just a sanity check on this line, not sure what this would do, would it fail silently if parent::setNewRevision() isn't implemented?

kalpaitch: It's always implemented, it's part of the core ContentEntityBase.

stemount: Was just testing you @kalpaitch

PR #131, opened 2018-01-06
PAR-1034: Add Tota11y web accessibility tool to site

stemount:
widget enabled for devs to check accessibility as well.


tests/src/features/0-user-journey-primary-authority-new-direct-partnership.feature
```diff
@@ -1,10 +1,13 @@
-@ci
 Feature: New Direct Partnership For PA To Approve
 
+    @ci @PAR1034
     Scenario: New Direct Partnership
+        Given I open the url "/user/login"
+        And I run tota11y against the current page
 
+    @ci
+    Scenario: New Direct Partnership
```
jon-edery: Not sure we should be adding tests for a dev tool to be part of the regression tests.  Can we check what the testing policy should be here.

kalpaitch: Agreed, it should be able to run as a separate set of tests.

jon-edery:
Apart from the issue around testing looks gd 👍

tests/src/features/0-user-journey-primary-authority-new-direct-partnership.feature
jaffamonkey: Will omit the step for now. Can run as separate job.

PR #135, opened 2018-01-10
PAR-1084 - Allow filtering partnerships by revoked status

stemount:
rd_help_desk___dashboard___primary_authority_register_and_2__stemount_winning____projects_par-beta__zsh_

Updating to the PAR Status Filter as per uniform with other views.

### DIFF
--- a/sync/views.view.helpdesk_dashboard.yml
+++ b/sync/views.view.helpdesk_dashboard.yml
@@ -596,70 +596,6 @@ display:
             authority_name: authority_name
             organisation_name: organisation_name
           plugin_id: combine
-        partnership_status:
-          id: partnership_status
-          table: par_partnerships_field_revision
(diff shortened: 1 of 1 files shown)


sync/views.view.helpdesk_dashboard.yml
```diff
@@ -622,67 +660,53 @@ display:
...
-                value: confirmed_business
-              4:
-                title: Active
-                operator: '='
-                value: confirmed_rd
-              5:
-                title: Revoked
-                operator: '='
-                value: revoked
+            group_items: {  }
```
kalpaitch: Why is group_items empty?

stemount: @kalpaitch the filter is changed from a grouped filter to our implementation of a parfilter thing.

PR #138, opened 2018-01-11
PAR-1044

jon-edery:
This ticket refers to a spike to investigate the ways in which we can implement the custom business logic rules required around revoked entities.

There instances of internal system queries that currently will return revoked partnership entities.

screen shot 2018-01-11 at 14 47 24

The internal system entities queries utilise the function getEntitiesByQuery() for retrieving par_data entities to be used for custom logic within the system.

The code in this branch extends the getEntitiesByQuery() providing a way to remove deleted and revoked entities from being retrieved and used internally, preventing various system failures.

There is an optional override parameter which will return revoked entities when required by bypassing this logic if needed.

There is also a need to address views that currently list partnership entities and possibly a need to extend the storage class for par_data entities for future requirements.

### DIFF
--- a/web/modules/custom/par_data/src/Entity/ParDataEnforcementAction.php
+++ b/web/modules/custom/par_data/src/Entity/ParDataEnforcementAction.php
@@ -120,7 +120,7 @@ class ParDataEnforcementAction extends ParDataEntity {
   }
 
   /**
-   * Check if this entity is revoked.
+   * Check if this entity is blocked.
    *
    * @return bool
    */
--- a/web/modules/custom/par_data/src/ParDataManager.php
+++ b/web/modules/custom/par_data/src/ParDataManager.php
@@ -586,7 +586,7 @@ class ParDataManager implements ParDataManagerInterface {
     $account = \Drupal::currentUser();
 
     // Remove revoked partnerships from internal system queries.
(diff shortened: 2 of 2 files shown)


stemount:
Good start @jon-edery and thanks for demoing to @kalpaitch and I yesterday, I have put this in the "Closed PR"s as a _WIP_ label but we should continue our conversation about the direction of this today now be able to discuss alternatives and pro/cons of using the many methods we can implement for this.

stemount:
So this ticket as we have discussed has quite a few ways of doing it, it's mostly just choosing the most slick one is the hardest bit and building that!

I think this is on the right track, we should definitely target either the entity load or some sort of query function so the change is more site-wide and not custom to any form.

We only really use this very quickly written function `getEntitiesByType` for things like our faux fuzzy matching contacts search where we need to make a query that does AND/OR.

In the `par_data.par_data_partnership.yml` file with the other entity properties, we could have a key that was which boolean states to show by default maybe?

```
default_show_states:
  - revoked
  - deleted
  - archived
```

then override the entity load function and check the entity type config to see if it is in a state it should be shown to Joe Public (e.g. not Helpdesk) - This idea does fall down a little bit if you wanted to load the entity with $blah->load();

> There is also a need to address views that currently list partnership entities
Aye. Maybe we could make a custom filter based upon the entity config as above. `PAR Published` or something!

> and possibly a need to extend the storage class for par_data entities for future requirements.
100%, soft delete/archive will be stories in the next few sprints I imagine.

PR #148, opened 2018-01-18
PAR-1109: Address field needs to accept non-UK postcodes.

stemount:
After trying quite a number of quite hacky methods I found that you can configure the constraints field level just by editing the field settings.

This PR also fixes the bug in PAR-996: Address validation error message format not GDS compliant - but not the titles.

- [x] update the file to not take into account State/County/Emirate and City/Town
- [x] confirm uk postcode checker works
- [x] add other to Nation
- [x] add a country select box if select Nation = Other

### DIFF
--- a/tests/src/features/2-user-journey-organisation-user-update-partnership.feature
+++ b/tests/src/features/2-user-journey-organisation-user-update-partnership.feature
@@ -18,7 +18,7 @@ Feature: Business User - Manage Addresses
 
         # EDIT REGISTERED ADDRESS
 
-        When  I click on the link "edit address"
+        When I click on the link "edit address"
         And I clear the inputfield "#edit-address-line1"
         And I clear the inputfield "#edit-address-line2"
         And I clear the inputfield "#edit-town-city"
@@ -31,12 +31,10 @@ Feature: Business User - Manage Addresses
         Then I expect that element ".error-summary" does exist
         And I add "1 Change St" to the inputfield "#edit-address-line1"
         And I click on the button "#edit-save"
-        Then I expect that element ".error-summary" does exist
+        When I click on the link "edit address"
         And I add "New Change" to the inputfield "#edit-address-line2"
-        And I click on the button "#edit-save"
-        Then I expect that element ".error-summary" does exist
-        When I add "London" to the inputfield "#edit-town-city"
-        When I add "London" to the inputfield "#edit-county"
+        And I add "London" to the inputfield "#edit-town-city"
+        And I add "London" to the inputfield "#edit-county"
         And I select the option with the text "England" for element "#edit-nation"
         When I click on the button "#edit-save"
         Then I expect that element "#edit-registered-address" contains the text "1 Change St"
--- a/tests/src/features/2-user-journey-organisation-user-update-partnership.feature
+++ b/tests/src/features/2-user-journey-organisation-user-update-partnership.feature
@@ -37,7 +37,7 @@ Feature: Business User - Manage Addresses
         Then I expect that element ".error-summary" does exist
         When I add "London" to the inputfield "#edit-town-city"
         When I add "London" to the inputfield "#edit-county"
-        And I select the option with the text "England" for element "#edit-country"
+        And I select the option with the text "England" for element "#edit-nation"
         When I click on the button "#edit-save"
         Then I expect that element "#edit-registered-address" contains the text "1 Change St"
         And I expect that element "#edit-registered-address" contains the text "New Change"
--- a/web/modules/custom/par_flows/src/Form/ParBaseForm.php
+++ b/web/modules/custom/par_flows/src/Form/ParBaseForm.php
@@ -446,7 +446,7 @@ abstract class ParBaseForm extends FormBase implements ParBaseInterface {
         else {
           $arguments = ['@field' => $field_label];
         }
-        $message = $this->t($violation->getMessage()->getUntranslatedString(), ['@field' => $field_label]);
+        $message = $this->t($violation->getMessage()->render(), ['@field' => $field_label]);
 
         $url = Url::fromUri('internal:#', $options);
         $link = Link::fromTextAndUrl($message, $url)->toString();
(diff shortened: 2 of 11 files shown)


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAddressForm.php
```diff
@@ -92,6 +96,7 @@ public function retrieveEditableValues(ParDataPartnership $par_data_partnership
       $this->loadDataValue("town_city", $address->get('locality')->getString());
       $this->loadDataValue("county", $address->get('administrative_area')->getString());
       $this->loadDataValue("country", $par_data_premises->get('nation')->getString());
+      $this->loadDataValue("country_code", $par_data_premises->get('country_code')->getString());
```
kalpaitch: There's no such field as this.

kalpaitch: I think you mean `$address->get('country_code')`

stemount: this is what chokes

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAddressForm.php
```diff
@@ -133,25 +138,36 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
       '#default_value' => $this->getDefaultValues("county"),
     ];
 
-    $form['country'] = [
+    $form['nation'] = [
       '#type' => 'select',
       '#title' => $this->t('Select your Nation'),
       '#options' => $premises_bundle->getAllowedValues('nation'),
       '#default_value' => $this->getDefaultValues("country"),
```
kalpaitch: This needs to change to `$this->getDefaultValues("nation")`

jon-edery:
Checks tests

tests/src/features/2-user-journey-organisation-user-update-partnership.feature
```diff
@@ -31,12 +31,10 @@ Feature: Business User - Manage Addresses
...
         And I add "1 Change St" to the inputfield "#edit-address-line1"
         And I click on the button "#edit-save"
-        Then I expect that element ".error-summary" does exist
+        When I click on the link "edit address"
         And I add "New Change" to the inputfield "#edit-address-line2"
-        And I click on the button "#edit-save"
-        Then I expect that element ".error-summary" does exist
-        When I add "London" to the inputfield "#edit-town-city"
-        When I add "London" to the inputfield "#edit-county"
+        And I add "London" to the inputfield "#edit-town-city"
```
kalpaitch: If we're changing these tests it seems we've done too much. Why is there an error here?

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAddressForm.php
```diff
@@ -188,8 +204,15 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
         'postal_code' => $this->getTempDataValue('postcode'),
       ];
 
+      $nation = $this->getTempDataValue('nation');
+
+      // Override any country code where nation field does not contain Other.
```
kalpaitch: Just a little note about perhaps getting this into the validation. If nation != other and country_code != GB throw a tizzy.

web/modules/custom/par_flows/src/Form/ParBaseForm.php
```diff
@@ -446,7 +446,7 @@ public function setFieldViolations($name, FormStateInterface &$form_state, Entit
         else {
           $arguments = ['@field' => $field_label];
         }
-        $message = $this->t($violation->getMessage()->getUntranslatedString(), ['@field' => $field_label]);
+        $message = $this->t($violation->getMessage()->render(), ['@field' => $field_label]);
```
kalpaitch: We didn't mean to change this.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAddressForm.php
```diff
@@ -188,8 +204,10 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
         'postal_code' => $this->getTempDataValue('postcode'),
       ];
 
+      $nation = $this->getTempDataValue('nation');
+
       $premises->set('address', $address);
-      $premises->set('nation', $this->getTempDataValue('country'));
+      $premises->set('nation', $nation);
```
kalpaitch:
```
if ($premises->get('address')->get('country_code') == GB) {
  $premises->set('nation', $nation);
}
```

stemount: @kalpaitch Other is a valid nation option so we should save it.

web/modules/custom/par_flows/src/Form/ParBaseForm.php
stemount: This was intentional, the errors do not even specify what field is wrong in the messages title. If this is still acceptable, then revert it.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAddressForm.php
stemount: `visible` is a show/hide in browser screen state only so it will still save a value. :)

web/modules/custom/par_flows/src/Form/ParBaseForm.php
kalpaitch: Hmm, without thoroughly testing I'm not sure about this. Would be good to keep in a separate ticket, perhaps we can put it in the ticket that was about correcting the validation, that way we can make sure we test thoroughly.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAddressForm.php
kalpaitch: Cool, I don't reckon we should have 'Other' as an option, we're entering data that isn't valid.

PR #149, opened 2018-01-18
Formz: Separation of form handling from controllers

kalpaitch:
* The main changes are that the negotiation/choosing of which flow is 'active' is now moved to a separate service, this allows us to work it out outside for the form controller.

* The handling of temporary stored data is also moved to a separate service which again allows us to get data stored for other locations (such as in the new form components)...

And the more controversial change is the addition of new form components which hopefully will allow us to build the form elements as plugins and reuse in different controllers without having to do any of that if flow = this then do x. Should separate the form elements from the logic of the pages. The only real thing about this work is it allows us to break up the form controllers into more manageable chunks (and keep them in features for each journey). This is the real questionable bit.

There is also a question that I've been unable to answer about whether we would ever have been able to do the kinda stuff we're doing in webforms, I suspect not, but obviously it's not ideal having a massively complex custom form building system.


web/modules/custom/par_flows/src/ParBaseInterface.php
```diff
@@ -18,33 +18,25 @@ public function getLoggerChannel();
...
-   * Get the current flow entity.
+   * Returns the PAR data manager.
    *
-   * @return \Drupal\Core\Entity\EntityInterface
-   *   The flow entity.
+   * @return \Drupal\par_flows\ParFlowDataHandlerInterface
+   *   Get the flow data handler.
    */
-  public function getFlow();
+  public function getflowDataHandler();
```
stemount: there are uses of `getflowDataHandler` and `getFlowDataHandler` which may help fix the tests.

kalpaitch: Yea need to look at tests. Nay wozza.

kalpaitch:
ha, yea still WIP. hopefully gonna get the new application journey done as an example this evening then we can start using that as the way forward.

PR #150, opened 2018-01-22
[PAR-1111] Use partnership entity to store legal entity references.

stemount:
- [x] Legal
entities should live on the partnership, changing form submit hander.
- [x] Retrieve legal entities from partnership
- [x] helper functions for partnership legal entities
- [x] Legal entities update to migrate these from the org to partnership

### DIFF
--- a/web/modules/custom/par_data/src/Entity/ParDataPartnership.php
+++ b/web/modules/custom/par_data/src/Entity/ParDataPartnership.php
@@ -270,6 +270,29 @@ class ParDataPartnership extends ParDataEntity {
   }
 
   /**
+   * Get legal entities for this partnership.
+   */
+  public function getLegalEntity() {
+    return $this->get('field_legal_entity')->referencedEntities();
+  }
+
+  /**
+   * Add a legal entity to partnership.
+   *
+   * @param ParDataLegalEntity $legal_entity
+   *   A PAR Legal Entity to add.
+   */
+  public function addLegalEntity(ParDataLegalEntity $legal_entity) {
+    // Retrieve existing legal entities.
+    $legal_entities = $this->getLegalEntity();
+
+    // Append new legal entity to existing entities.
+    $legal_entities[] = $legal_entity;
+
+    $this->set('field_legal_entity', $legal_entities);
+  }
+
+  /**
    * {@inheritdoc}
    */
   public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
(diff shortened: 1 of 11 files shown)


web/modules/custom/par_data/src/Entity/ParDataPartnership.php
```diff
@@ -269,6 +269,29 @@ public function isCoordinated() {
...
+    return $this->get('field_legal_entity')->referencedEntities();
+  }
+
+  /**
+   * Add a legal entity to partnership.
+   *
+   * @param ParDataLegalEntity $legal_entity
+   *   A PAR Legal Entity to add.
+   */
+  public function addLegalEntity(ParDataLegalEntity $legal_entity) {
```
kalpaitch: Don't need this, there's such a method as `append` on FieldItemListInterfaces.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsLegalEntityForm.php
```diff
@@ -180,12 +181,18 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
       ]);
       $legal_entity->save();
 
-      // Now add the legal entity to the organisation.
+      // Now add the legal entity to the partnership.
       $par_data_partnership = $this->getRouteParam('par_data_partnership');
-      $par_data_organisation = current($par_data_partnership->getOrganisation());
+      $par_data_partnership->addLegalEntity($legal_entity);
```
kalpaitch: See note about `appendItem()`

kalpaitch:
Will review again once https://github.com/UKGovernmentBEIS/beis-primary-authority-register/pull/148 is updated.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsLegalEntityForm.php
stemount: Yes, this is actually uniform across the board… see 
par-beta____projects_par-beta__-_____web_modules_features_par_dashboards_src_controller_pardashboardsdashboardcontroller_php__par-beta_

kalpaitch: Cool, it's highly possible that we weren't aware of `appendItem` at the time but now is a good chance to update that.

stemount:
Yes, let's totally clean that up in its own PR @kalpaitch I much prefer the policy of refactoring in a seperate PR. With these new flow plugins we can get rid of cruft as we go.

PR #151, opened 2018-01-22
Update business application flow branch with latest master

jon-edery:


kalpaitch:
Let's do this when the last little comments on the address ticket are updated - https://github.com/UKGovernmentBEIS/beis-primary-authority-register/pull/148#pullrequestreview-90809207

jon-edery:
it just means we can keep current PR's cleaner.  while avoiding large merge conflicts I would prefer to do this more frequently

stemount:
Makes sense @jon-edery

PR #152, opened 2018-01-22
PAR-1105: Switch organisation details link based upon partnership status

jon-edery:
Mapping to journey contextual rules (based on the partnership application state) :

Confirmed partnerships map apply journey.
Unconfirmed map to the (old) update journey

### DIFF
--- a/web/modules/custom/par_data/src/Views/ParDataViewsData.php
+++ b/web/modules/custom/par_data/src/Views/ParDataViewsData.php
@@ -78,6 +78,16 @@ class ParDataViewsData extends EntityViewsData implements EntityViewsDataInterfa
       ],
     ];
 
+    // PAR Partnership Flow Link.
+    $data[$this->entityType->getDataTable()]['par_partnership_flow_link'] = [
+      'title' => t('PAR Partnership Flow Link'),
+      'field' => [
+        'title' => t('PAR Partnership Flow Link'),
+        'help' => t('Provides a link to the relevant partnership journey determined by the partnership status.'),
+        'id' => 'par_partnership_flow_link',
+      ],
+    ];
+
     // Custom filter for Par Membership checks.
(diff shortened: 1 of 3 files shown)


kalpaitch:
Just as a thought, we can do this entire thing without writing any code:
screenshot from 2018-01-23 13-13-33
screenshot from 2018-01-23 13-11-30

stemount:
Let's bash heads together on this tomorrow it may be valid as a plugin but lets see how well this no results thing is as that would be a winner.

jon-edery:
The config option would work for access control but would not factor in the partnership state. Personally

stemount:
This PR has been merged and tested in conjunction w/ PAR-1104.

stemount:
RE: above comments about empty links, partnership status etc…

3 Par Flow Links and create a custom field with the output of all of them?

{{par_link_1}} /authority-details (if active and PA, allow)
{{par_link_2}} /organisation-details (if partnership active, allow)
{{par_link_3}} /new-apply-journey/whatever/url (if partnership pending, allow this link)

then modify the access callback to controlling if these links are hidden or not.

This would get rid of the double links-gate.

primary_authority_register___pending_applications___primary_authority_register

web/modules/custom/par_data/src/Views/ParDataViewsData.php
```diff
@@ -78,6 +78,16 @@ public function getViewsData() {
       ],
     ];
 
+    // PAR Partnership Flow Link.
+    $data[$this->entityType->getDataTable()]['par_partnership_flow_link'] = [
```
kalpaitch: Should conditionally apply this to partnership entities only.

kalpaitch:
Closing as this is now incorporated into PR https://github.com/UKGovernmentBEIS/beis-primary-authority-register/pull/158

PR #158, opened 2018-01-24
PAR-1104: See pending applications view.

stemount:
This branch has merged PAR-1105 in to maintain the fields.

primary_authority_register___primary_authority_register

primary_authority_register___pending_applications___primary_authority_register

### DIFF
--- a/web/modules/custom/par_data/src/Plugin/views/field/ParPartnershipFlowLink.php
+++ b/web/modules/custom/par_data/src/Plugin/views/field/ParPartnershipFlowLink.php
@@ -1,0 +1,109 @@
+<?php
+
+/**
+ * @file
+ * Definition of Drupal\par_data\Plugin\views\field\ParPartnershipsCombinedStatusField
+ */
+
+namespace Drupal\par_data\Plugin\views\field;
+
+use Drupal\Component\Utility\Html;
+use Drupal\Core\Link;
+use Drupal\Core\Url;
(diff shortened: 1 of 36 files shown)


web/modules/custom/par_data/src/Views/ParDataViewsData.php
```diff
@@ -78,6 +78,16 @@ public function getViewsData() {
       ],
     ];
 
+    // PAR Partnership Flow Link.
+    $data[$this->entityType->getDataTable()]['par_partnership_flow_link'] = [
```
kalpaitch: If we do this we should conditionally add to only the partnership entity.

web/modules/custom/par_data/src/Plugin/views/field/ParPartnershipFlowLink.php
```diff
@@ -0,0 +1,109 @@
...
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Field handler to get the PAR Partnership journey links.
+ *
+ * @ingroup views_field_handlers
+ *
+ * @ViewsField("par_partnership_flow_link")
+ */
+class ParPartnershipFlowLink extends FieldPluginBase {
```
kalpaitch: Still leaning on thinkin we should probably try to achieve this without writing a new views field specifically for partnerhips.

PR #160, opened 2018-01-25
oops, we updb way before the site is actually properly configured/setup.

stemount:
This causes an issue because the legal entity field is required for update_8019… so we should get this into master asap.

### DIFF
--- a/drupal-update.sh
+++ b/drupal-update.sh
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
 ## Commands that must be run to update a drupal instance.
 ## Use as `sh ./drupal-update.sh /var/www/html`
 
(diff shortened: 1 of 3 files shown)


drupal-update.sh
```diff
@@ -18,15 +19,16 @@ cd ${ROOT}/web; ../vendor/drush/drush/drush sset system.maintenance_mode 1;
...
-cd ${ROOT}/web; ../vendor/drush/drush/drush updb -y;
 # Import configuration twice to fix a problem with config import when new modules are added to 'core.extensions.yml'.
 cd ${ROOT}/web; ../vendor/drush/drush/drush cim -y; ../vendor/drush/drush/drush cim -y;
 # To doubly make sure drush registers features commands.
 cd ${ROOT}/web; ../vendor/drush/drush/drush cc drush;
 # Revert all features
 cd ${ROOT}/web; ../vendor/drush/drush/drush fra -y;
 
+# Run db updates.
+cd ${ROOT}/web; ../vendor/drush/drush/drush updb -y;
```
kalpaitch: Yea I thought about this, btw I think we always need to do an `cim` after an `updb`, had this issues previously, and we need to do both before the `fra`, reluctant to do this but think it may be the only option.

web/sites/default/settings.local.travis.php
```diff
@@ -23,7 +23,9 @@
 
 // Ensure travis always runs with the same memory that other environments do.
 ini_set('memory_limit', '2048M');
-ini_set('max_execution_time', 300);
+
+// Set 10min execution time as some drush updb commands are CPU time expensive.
+ini_set('max_execution_time', 600);
```
kalpaitch: drush cli should be set with infinite execution time and memory, will ask @chris-moreton what he thinks.

kalpaitch: Ignore this, probably not possible to configure Travis php cli.

stemount:
This PR needs a couple of deployments.

`We're going to release v13.1.0 because we need to do a two step release for migrating the legal entities to partnerships.`

PR #161, opened 2018-01-25
Repatch sql-dump fix

chris-moreton:
Reapplies the sql-dump fix. Have tested the sql-dump command and it seems to be working without problems.

To test, ensure that you have no unpushed commits and rebuild the vagrant environment and see if it works by rebuilding from the branch, i.e.

sh {{REPO_DIRECTORY}}/vagrant.sh PAR-1094


kalpaitch:
Just need to check that `sql-dump` dumps the database without ownership and with empty cache tables, and that `sql-drop` does what it says on the tin.

jaffamonkey:
Working ok from my more limited perspective (i.e. vagrant build up and running ok and tests passing)

kalpaitch:
Drop tables isn't dropping with this patch (on my machine anyway), also don't know why latest master has failed the tests.

PR #166, opened 2018-01-26
Content updates: Sam's PR

kalpaitch:


kalpaitch:
@samlionheart If we could just make sure the text changes on these titles works it's way back into the UX wireframes.

PR #167, opened 2018-01-29
PAR-1094 

jon-edery:
This PR adds a new event subscriber for the nominated (approved) partnership.  

this event needs testing via email notification service on staging.


web/modules/custom/par_data/src/Event/ParDataEvent.php
```diff
@@ -37,6 +37,15 @@ class ParDataEvent extends Event implements ParDataEventInterface {
...
   const DELETE = 'par_data.entity.delete';
 
+  /**
+   * The name of the event triggered when an existing par entity is approved.
+   *
+   * @Event
+   *
+   * @var string
+   */
+  const APPROVED = 'par_data.entity.approve';
```
kalpaitch: If we're going to have an event for each status change we should probably get the names from the `ParDataTypeInterface::getAllowedValues(PatDataTypeInterface::getConfigurationElementByType('entity', 'status_field'))` rather than from a constant here.

web/modules/features/par_rd_help_desk_flows/src/Form/ParRdHelpDeskApproveConfirmForm.php
```diff
@@ -165,6 +166,12 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
         $this->getLogger($this->getLoggerChannel())
           ->error($message, $replacements);
       }
+      else {
```
kalpaitch: Thinking we should only really be setting entity events in either the 1) ParDataEntity class or the 2) ParDataStorage class.

1) If worried about the fact that `setParStatus` doesn't save the entity we could change this, it would probably make more sense.
2) Otherwise it would be a perfectly legitimate thing to fire the events in `ParDataStorage::save` and just do some logic to check if the status field is being changed.

jon-edery:
Comments have been addressed this feature needs testing in an environment that is setup for email notifications (staging).

kalpaitch:
Closing as duplicate of PAR-1099 PR

PR #172, opened 2018-01-30
PAR-803: Moved saving of form data to the end.

kalpaitch:


kalpaitch:
This is failing because the tests are expecting the update journey but have now finally configured the application journey. Waiting on new tests for the application journey.

PR #177, opened 2018-02-01
Content updates

kalpaitch:


kalpaitch:
Will probably need to update titles and will need a dev to revert features @jon-edery @stemount @kalpaitch

PR #181, opened 2018-02-05
PAR-1112

jon-edery:
This PR ensures deleted partnership entities are removed from the system.  An additional update hook to removed partnership entities that have been sort deleted is being tested will be added.


jon-edery:
Tested both partnership hard deletes removes partnership entities from the database with no regressions to the sort delete functionality.

PR #182, opened 2018-02-06
Update hook to remove soft deleted partnerships

jon-edery:


jon-edery:
I would recommend a full site regression automated testing instance executed on a production cloned database to verify that removing these partnerships will not have an impacted the site in an unforeseen way.

kalpaitch:
Sorry @jon-edery Didn't see the A/C on the ticket, I'm being naughty and going with what I thought the A/C were. I've put a comment on the ticket now, we need to update A/C and look at this ticket more thoroughly, have dropped the ticket out of the sprint we'll need to discuss this more I think.

PR #184, opened 2018-02-08
PAR-1112: Updated views to not show soft deleted entities.

kalpaitch:
This is a temporary measure while we contemplate the implications of removing all these entities permanently.


sync/views.view.par_user_enforcement_list.yml
```diff
@@ -935,6 +973,44 @@ display:
...
+            widget: select
+            multiple: false
+            remember: false
+            default_group: All
+            default_group_multiple: {  }
+            group_items: {  }
+          entity_type: par_data_enforcement_notice
+          entity_field: deleted
+          plugin_id: boolean
       filter_groups:
```
jon-edery: Looks like the same filter has been added 2 times here.

jon-edery: This maybe just a result of multiple displays on the view?

jon-edery:
Looks gd.  would just confirm we are not adding the same filter 2 times on a the view and we are happy for the vbo export config to be added here.

sync/views.view.par_user_enforcement_list.yml
kalpaitch: Yea multiple displays. There is some reason whey the two displays had to use slightly different fields and therefore add the deleted field twice.

composer.json
```diff
@@ -49,6 +49,7 @@
         "drupal/twig_field_value": "~1.1.0",
         "drupal/twig_tweak": "~1.8.0",
         "drupal/ultimate_cron": "^2.0@alpha",
+        "drupal/vbo_export": "^2.0",
```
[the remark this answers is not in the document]
kalpaitch: Kinda trying to sneak it in yea :)

PR #193, opened 2018-02-12
PAR-1128, PAR-1127, PAR-1129

jon-edery:
This PR contains code for 3 of the tickets noted above.   The listing page is not complete as we need other tickets to merge in additional features for this page to be completed.


sync/par_flows.par_flow.partnership_coordinated.yml
```diff
@@ -119,3 +120,5 @@ steps:
       cancel: 1
   17:
     route: view.par_user_partnerships.partnerships_page
+  18:
```
kalpaitch: Let's re-order this so that the last step is actually the last step.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsDetailsForm.php
```diff
@@ -78,11 +78,24 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
...
-      $form['members'] = $this->renderSection('Members', $par_data_partnership, ['field_coordinated_business' => 'title']);
+      $form['members_link'] = [
+        '#type' => 'fieldset',
+        '#title' => t('Number of members'),
+        '#attributes' => ['class' => 'form-group'],
+        '#collapsible' => FALSE,
+        '#collapsed' => FALSE,
+      ];
+
+      $form['members_link']['link'] = [
```
kalpaitch: Just a note that we should try to avoid be merging this with the link until everything else is done. We can split the link back into it's own ticket so that we can get everything else in master.

stemount:
Is this PR ready?

PR #195, opened 2018-02-13
PAR-1167 - Upload confirmation page (Uploaded csv data processed and added to the queue).

baberabbasi:


web/modules/features/par_member_upload_flows/src/Plugin/QueueWorker/MemberUploadQueue.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+/**
+ * @file
+ * Contains \Drupal\par_member_upload_flows\Plugin\QueueWorker\MemberUploadQueue.
+ */
+
+namespace Drupal\par_member_upload_flows\Plugin\QueueWorker;
+
+use Drupal\Core\Queue\QueueWorkerBase;
```
kalpaitch: Just a note for @stemount @baberabbasi:

Jon and I were discussing possible ways to determine when a particular queue is done processing, and I think the only way is to have one queue per upload. We can try to create dynamic queues.

web/modules/features/par_member_upload_flows/src/Form/ParMemberConfirmUploadFlowsForm.php
```diff
@@ -0,0 +1,69 @@
...
+
+    // Load temporary csv data and assign it to a variable.
+    $csv_data = $this->getFlowDataHandler()->getTempDataValue('coordinated_members', $cid);
+
+    // Create CRON queue with a unique name.
+    $queue = \Drupal::queue('par_member_upload');
+
+    // Create CRON QUEUE item which is added to the queue and will be
+    // triggered next time when CRON executes.
+    $queue->createItem($csv_data);
```
kalpaitch: Let's create a queue item based on one $csv_data row at a time.

```
foreach ($csv_data as $row) {
  $queue->createItem($row);
}
```

We'll probably need to change this once we've worked everything else out but this gives us a good base.

PR #197, opened 2018-02-13
PAR-1131: Created new member add journey.

kalpaitch:
Titles are a little bit changed because they're sometimes inconsistent from one page to another.

screenshot from 2018-02-13 16-41-40


stemount:
bug found with legal entities having empty select: 

messages image 1400268007

address field error message is maybe too ambiguous e.g. postcode was not entered in this instance so no red marking on the correct field that needs addressing.

messages image 2639792185

stemount:
Seems to go through nicely like a journey, start date and covered by inspection withstanding.

PR #203, opened 2018-02-16
Tesco accounts potential fix WIP

stemount:
This requires thorough testing, lets chat about ways to test this properly

I used

`    $par_data_manager = \Drupal::service('par_data.manager');

    $users = User::loadMultiple();

    foreach (array_slice($users, 200, 10) as $id => $user) {
      var_dump(array_keys($par_data_manager->hasMemberships($user)['par_data_partnership']));
    }
`

to sense check this by checking 10 at a time…

### DIFF
--- a/web/modules/custom/par_data/src/ParDataManager.php
+++ b/web/modules/custom/par_data/src/ParDataManager.php
@@ -311,11 +311,6 @@ class ParDataManager implements ParDataManagerInterface {
       }
 
       foreach ($referenced_entities as $entity_id => $referenced_entity) {
-        // Always skip lookup of relationships for people.
-        if ($referenced_entity->getEntityTypeId() === 'par_data_person') {
-          continue;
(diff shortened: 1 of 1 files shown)


kalpaitch:
I think this could be potentially dangerous. It would allow any person record discovered on another entity, partnership etc to be included.

kalpaitch:
I think it would be good to find out why this issue is occurring before fixing as this is a potentially troublesome area. What is the difference (in reference links) between the tesco partnership that is showing, and the one that is not.

stemount:
It is worth noting on this issue that https://regulatorydelivery.atlassian.net/browse/PAR-1194 applies as well.

kalpaitch:
True, we potentially need to work this all out before we release members.

kalpaitch:
Closing while further investigation underway.

PR #207, opened 2018-02-19
PAR-1197  minor config update

jon-edery:


jon-edery:
Good idea have created a new filter that uses the internal status config rather then hard coded values as it was built previously.

sync/views.view.par_user_partnerships.yml
```diff
@@ -681,7 +678,7 @@ display:
                 operator: '='
                 value: confirmed_authority
               3:
-                title: 'Confirmed by Business'
+                title: 'Confirmed by Organisation'
```
kalpaitch: We should probably get this filter off the 'default' view config but given that it's not used it's hardly important right now.

sync/views.view.par_user_partnerships.yml
```diff
@@ -1362,6 +1359,343 @@ display:
       display_extenders: {  }
       path: partnerships
       display_description: ''
+      filters:
```
kalpaitch: Can't see the existing filter as having been removed, will import and check in a mo.

stemount:
When all traces of the old filter gone then please feel free to dismiss my Request changes review 💯

PR #209, opened 2018-02-19
PAR-1190: Resolving bug with adding new contacts (and advice)

kalpaitch:
The bug here was the fact that we'd moved from using the form id as a cache key for the form data to the route name.

We'd done this so that we could **avoid issues where the same form existed twice in a flow** (contact adding and editing). Previously the same cache was used for both forms, but now the data was only picked up from one of these, meaning that editing a contact worked, but not adding a contact.

The solution here is to use **individual form_ids** (unique for the flow) for the same form controller (contact_add and contact_edit), this has required a bigger shift towards having abstracted `form_data` keys that allow the current step to dictate which contact form it wants to get data form.

In this example the contact suggestion form is related to the contact add form but not the contact edit form, we can set this relationship using the `form_data` keys.

## Review notes:
In summary the [form id is now read from the `form_id` value in the flow step configuration](https://github.com/UKGovernmentBEIS/beis-primary-authority-register/pull/209/files#diff-77079c597435983accd9a5ca0200dcd2R171). This means that we can use the same form but a different form_id.

Most of the files in this PR are either [removing the hard coded `form_id`](https://github.com/UKGovernmentBEIS/beis-primary-authority-register/pull/209/files#diff-3a33efc7473622c704a4c32eea3b8838L39) or [changing the `form_id`s in the flow step configuration to be more unique](https://github.com/UKGovernmentBEIS/beis-primary-authority-register/pull/209/files#diff-20ca9e4c8b27cd0cd2bc3a529965fb85R66).

### DIFF
--- a/drupal-update.sh
+++ b/drupal-update.sh
@@ -5,8 +5,7 @@
 if [ -n "$1" ]; then
   ROOT=$1
 else
-  echo "Must pass the project root as the first argument";
-  exit 1;
+  ROOT=$PWD
 fi
 
 echo "Current working directory is ${ROOT}/web"
--- a/sync/par_flows.par_flow.partnership_authority.yml
+++ b/sync/par_flows.par_flow.partnership_authority.yml
@@ -3,7 +3,7 @@ langcode: en
 status: true
 dependencies: {  }
 _core:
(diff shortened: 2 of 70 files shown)


stemount:
Yea this is an important PR lets get it in

PR #211, opened 2018-02-20
PAR-1142: Member cease journey.

kalpaitch:


web/modules/custom/par_forms/src/Element/GdsDate.php
```diff
@@ -44,7 +44,6 @@ public function getInfo() {
         [$class, 'preRenderCompositeFormElement'],
       ],
       '#date_date_format' => 'Y-m-d',
-      '#theme_wrappers' => ['form_element'],
```
stemount: is this a mistake?

stemount:
Looks good, just a rogue line of code maybe?

web/modules/custom/par_forms/src/Element/GdsDate.php
kalpaitch: It was repeating the '#title' and description wrappers. Not sure why or if there's a better way.

stemount: No, I reckon this is not needed, preRenderCompositeFormElement adds a fieldset, so 👍

PR #212, opened 2018-02-21
PAR-1015: New update member journey.

kalpaitch:
This is incomplete, in that it doesn't currently save any data, but I wanted to get this PR in so that:
a) we can add the link to this journey to the member list
b) so that @jaffamonkey can start testing.

Given that the only change to existing code is 1 line (https://github.com/UKGovernmentBEIS/beis-primary-authority-register/compare/master...PAR-1015#diff-28f2f2a5f17d1a3159fddf43736036a9) I think it's probably safe enough to merge in if you're happy with this.


web/modules/features/par_member_update_flows/par_member_update_flows.module
```diff
@@ -0,0 +1,5 @@
+<?php
+
+/**
+ * @file
```
jon-edery: We may need to cut down on the level of documentation here.  Reading this comment block may take more then the acceptable amount of time a developer should be spending understanding the purpose of this module.

PR #230, opened 2018-02-22
PAR-1209: Some UX amends to member journeys.

kalpaitch:
Also includes links for remaining journeys: PAR-1140, PAR-1145

### DIFF
--- a/web/modules/custom/par_flows/src/Controller/ParBaseController.php
+++ b/web/modules/custom/par_flows/src/Controller/ParBaseController.php
@@ -6,6 +6,7 @@ use Drupal\Component\Plugin\PluginManagerInterface;
 use Drupal\Core\Cache\RefinableCacheableDependencyTrait;
 use Drupal\Core\Config\Entity\ConfigEntityStorageInterface;
 use Drupal\Core\Controller\ControllerBase;
+use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
 use Drupal\par_data\ParDataManagerInterface;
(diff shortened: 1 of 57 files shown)


web/modules/features/par_rd_help_desk_flows/src/Form/ParRdHelpDeskRevokeConfirmForm.php
```diff
@@ -30,7 +34,18 @@ public function titleCallback() {
...
    * {@inheritdoc}
    */
-  public function accessCallback(ParDataPartnership $par_data_partnership = NULL) {
+  public function accessCallback(Route $route, RouteMatchInterface $route_match, AccountInterface $account) {
+    try {
+      $this->getFlowNegotiator()->setRoute($route_match);
+      $this->getFlowDataHandler()->reset();
+      $this->loadData();
+    } catch (ParFlowException $e) {
+
```
jon-edery: Minor: should we log access failures to catch system errors here?

web/modules/custom/par_flows/src/Controller/ParBaseController.php
```diff
@@ -150,13 +152,20 @@ public function build($build) {
...
+   *
    * @see \Drupal\Core\Access\AccessResult
    *   The options for callback.
    *
    * @return \Drupal\Core\Access\AccessResult
    *   The access result.
    */
-  public function accessCallback() {
+  public function accessCallback(Route $route, RouteMatchInterface $route_match, AccountInterface $account) {
     return $this->accessResult ? $this->accessResult : AccessResult::allowed();
```
jon-edery: are the params here just for constancy we dont seem to be using them here in the access call bk or am I missing something?

jon-edery:
This looks done a few comments (minor) more questions then change requests.

web/modules/custom/par_flows/src/Controller/ParBaseController.php
kalpaitch: Yup, for consistency.

web/modules/features/par_rd_help_desk_flows/src/Form/ParRdHelpDeskRevokeConfirmForm.php
kalpaitch: Probably, will add as notices so that we don't flood sentry with them.

PR #282, opened 2018-05-25
PAR-1181, PAR-864: Updated enforcement notification emails.

kalpaitch:


web/modules/custom/par_data/par_data.module
```diff
@@ -48,57 +48,6 @@ function par_data_theme() {
...
- */
-function par_data_entity_insert(EntityInterface $entity) {
-  if ($entity instanceof ParDataEntityInterface) {
-    $event = new ParDataEvent($entity);
-    $dispatcher = \Drupal::service('event_dispatcher');
-    $dispatcher->dispatch(ParDataEvent::CREATE, $event);
-  }
-}
-
-/**
```
jon-edery: any reason why we are removing these events?

web/modules/custom/par_data/src/Entity/ParDataEnforcementNotice.php
```diff
@@ -271,6 +271,68 @@ public function approve() {
...
+    $raw_status = $this->getRawStatus();
+    return $field_name && $raw_status ? $this->getTypeEntity()->getAllowedFieldlabel($field_name, $raw_status) : '';
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function setParStatus($value) {
+    // This status is generated dynamically and cannot be saved.
+    return;
```
jon-edery: Not sure what the point in this function is?  is this for future dev?

web/modules/custom/par_data/par_data.module
kalpaitch: I figured it was probably better to use the same events that were basically already in a Drupal core patch - https://www.drupal.org/project/drupal/issues/2551893

web/modules/custom/par_data/src/Entity/ParDataEnforcementNotice.php
kalpaitch: We do use it for other entities, but because ENs don't have a status field I was planning to put the notification in here but then decided against that. Will remove it for sure.

PR #315, opened 2018-06-26
PAR-1297: Updated nation data.

kalpaitch:


web/modules/custom/par_data/par_data.install
```diff
@@ -390,3 +390,63 @@ function par_data_update_8026() {
   // Then run purge to make sure all other fields are gone.
   field_purge_batch(500);
 }
+
+/**
+ * Correct missing addresses.
+ */
+function par_data_update_8028() {
```
jon-edery: should be par_data_update_8027

PR #322, opened 2018-07-04
Add new environment manifests

chris-moreton:


chris-moreton:
@kalpaitch You're right - I need to add those settings files.

PR #393, opened 2019-01-30
PAR-1348: Existing organisation selection for partnerships.

kalpaitch:


web/modules/custom/par_data/par_data.routing.yml
```diff
@@ -32,3 +32,13 @@ par_data.reports_page:
     _permission: 'administer site configuration'
   options:
     _admin_route: FALSE
+
+par_data.references_list:
+  path: '/admin/reports/fields/par'
```
kalpaitch: @jon-edery There's a new page here which just helps view all the entity reference fields for par data in one place. Combined with the data model it's a helpful reference to understand what's going on.

PR #395, opened 2019-02-08
Make feature tests run on CircleCI

kalpaitch:
In summary:
* Full end-to-end tests have been added to the circleci config and now run on all circleci builds
* Utilising the approval mechanism in circleci we've now introduced the deployment script into the pipeline
* The deployment script has also been updated to be more robust, and also to spin up new instances (it used to only push to existing instances), it can still be run manually as before.
* Disabled travisci and removed the old vm, docker and any scripts that are no longer used in this new deployment pipeline.
* There also appears to be a fix for the issue mentioned in Birmingham about certain notifications causing errors - https://github.com/UKGovernmentBEIS/beis-primary-authority-register/pull/395/files#diff-79ff390bd22e290ca1a545921bebdc59


devops/docker/Dockerfile
```diff
@@ -1,40 +1,133 @@
...
+    && php -r "if (hash_file('sha384', 'composer-setup.php') === '48e3236262b34d30969dca3c37281b3b4bbe3221bda826ac6a9a62d6444cdb0dcd0615698a5cbe587c3f0fe57a54d8f5') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;" \
+    && php composer-setup.php \
+    && php -r "unlink('composer-setup.php');" \
+    && mv composer.phar /usr/local/bin/composer
+
+# Install node
+## Using the nvm installation from https://github.com/mklement0/n-install
+RUN cd /tmp \
+    && sh -c 'git clone https://github.com/tj/n && cd n && make install' \
     && n 7.2.1 \
```
kalpaitch: Ideally we'd update node to the latest LTS version.

PR #416, opened 2019-03-18
PAR-1158 external template links

jon-edery:
Screen Shot 2019-03-18 at 10 34 26


jon-edery:
Vars updated to conform with standards.  The link construction is a correct way of implementing this. Examples can be found in other modules in our code base.

PR #480, opened 2019-07-30
PAR-1438, PAR-1441 Partnership application cancellation journey

jon-edery:
This code defines the cancel application journey for un-nominated partnerships.  Access control is defined to ensure business rules are enforced.

Email notifications tested locally.
Access control logic tested.
Business rules enforced for partnership states tested.
Partnership data stored for par entities tested.
Views integration with new feature tested.
correct data stored against partnership entities within the DB tested.

Cancel a partnership application: 

Screen Shot 2019-08-06 at 09 54 30

Confirmation scree:

Screen Shot 2019-08-06 at 14 15 54

Help Desk Cancel link:

Screen Shot 2019-08-06 at 14 17 15


web/modules/custom/par_data/par_data.install
```diff
@@ -751,3 +751,38 @@ function par_data_update_8035() {
...
+ */
+function par_data_update_8036() {
+  $entity_definition_manager = \Drupal::service('entity.definition_update_manager');
+
+// New fields definitions to be added to the partnership entity.
+// @see https://www.drupal.org/node/3034742
+  $storage_definitions = [];
+
+// Partnership delete reason.
+  $storage_definitions['deleted_reason'] = BaseFieldDefinition::create('text_long')
```
kalpaitch: @jon-edery See comment on https://regulatorydelivery.atlassian.net/browse/PAR-1442

We talked about spiking/discussing this before going ahead with a particular solution. Hence why it was not in this sprint. Do you want to have a think about the three options...

sync/core.entity_view_display.par_data_advice.advice.default.yml
```diff
@@ -54,6 +54,13 @@ content:
     settings: {  }
     third_party_settings: {  }
     type: text_default
+  deleted_reason:
```
kalpaitch: What I'm confused about is if this field has been updated on all par entities how come I can only see the view_display...default.yml files for advice, partnerships and people. Shouldn't there be a bunch of them for all the other entities??

PR #504, opened 2019-09-09
PAR-1482: Updated reference field display.

kalpaitch:


web/modules/custom/par_data/src/Entity/ParDataPartnership.php
```diff
@@ -297,6 +293,9 @@ public function getAuthority($single = FALSE) {
    */
   public function getCoordinatedMember($single = FALSE) {
     $members = $this->get('field_coordinated_business')->referencedEntities();
+    $members = array_filter($members, function ($member) {
```
kalpaitch: @jon-edery What do you think, it seems ridiculous if we have to do this for all `referencedEntities()` calls, cos there is no way of hooking into this other than at the entity storage level (`loadMultiple`), which would screw all sorts up.

kalpaitch: My feeling is this might just be a temporary solution.

PR #524, opened 2019-10-23
Inspection update

kalpaitch:


sync/par_data.par_data_inspection_plan_type.inspection_plan.yml
```diff
@@ -14,14 +14,18 @@ isArchivable: false
...
+      current:
+        - default
+      expired:
+        - current
   inspection_status:
     allowed_values:
       current: Current
       expired: Expired
       in_progress: 'In Progress'
-      revoked: Revoked
```
jon-edery: Are we no longer revoking ?

kalpaitch: Yea, but this is one of those standard statuses, you don't need to add it to every entity.

PR #526, opened 2019-11-06
PAR-1497 support for setting a revoke timestamp on an entity date field.

jon-edery:
Added support for entities to be able to specify a date field that should be updated to be inline with the revoke date.  This is used in inspection plans for example when revoking we want to update the valid until date value to the revoke timestamp.  

The core revoke function has been extended to provide optional support for other entities to do the same.


web/modules/custom/par_data/src/Entity/ParDataEntity.php
```diff
@@ -284,10 +285,16 @@ public function delete($reason = '') {
...
+   *   The field name of the date field which a revoke timestamp should be applied.
+   *
+   * @param Array $date_rage_values
+   *   The array of the date range values to set.
+   *
    * @return boolean
    *   True if the entity was revoked, false for all other results.
    */
-  public function revoke($save = TRUE, $reason = '') {
+  public function revoke($save = TRUE, $reason = '', $entity_field = NULL, $date_rage_values = NULL) {
```
kalpaitch: IMO I think it's probably best not having the date setting logic in a generic way like this (on the `ParDataEntity` class):

a) It's a different field on partnerships and inspection plans.
b) it's only on these two entities and no others.
c) it adds a couple of unnecessary parameters which could lead to the method becoming a bit untidy and messy if extended further.

The way we currently have it I think the two options are:

1) to make all revokable entities use use a standard field for end dates (and migrate the partnership/inspection plan dates to this).
2) to override the revoke method on the `ParDataPartnership` and `ParDataInspectionPlan` classes and applies the extra parameters/date setting in these methods only.

I think 2 is probably what I'd reckon easiest and best at this stage, and resolves the current issue/bug, with a view to maybe review later.

jon-edery: Cool I have made the changes.

PR #527, opened 2019-11-15
Par 1499 inspection plan auto restore

jon-edery:
Prevent system loop hole for inspection plans being re-activated indirectly.

### DIFF
--- a/web/modules/custom/par_data/src/Entity/ParDataInspectionPlan.php
+++ b/web/modules/custom/par_data/src/Entity/ParDataInspectionPlan.php
@@ -146,9 +146,8 @@ class ParDataInspectionPlan extends ParDataEntity {
       // Set revoke reason.
       $this->set(ParDataEntity::REVOKE_REASON_FIELD, $reason);
 
-      // In the case that inspection plans are being revoked as a resulted of a partnership being
-      // revoked we want to prevent historical revoked ISP's from being auto activated in the case
-      // of a partnership being restored.
+      // If the inspection plan is being revoked as a the result of a partnership revocation
+      // keep the original revoke date so that the inspection plan can be restored later.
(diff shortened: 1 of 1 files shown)


web/modules/custom/par_data/src/Entity/ParDataInspectionPlan.php
```diff
@@ -145,14 +145,33 @@ public function revoke($save = TRUE, $reason = '') {
...
 
       return $save ? ($this->save() === SAVED_UPDATED || $this->save() === SAVED_NEW) : TRUE;
     }
     return FALSE;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function unrevoke($save = TRUE) {
```
kalpaitch: I think this looks good, we'll have to remember that if we want to unrevoke an inspection plan we will have to ask for (and set, separately) the new expiry date. Doesn't impact this though.

web/modules/custom/par_data/src/Entity/ParDataInspectionPlan.php
```diff
@@ -145,14 +145,33 @@ public function revoke($save = TRUE, $reason = '') {
       $this->setNewRevision(TRUE);
       // Set revoke reason.
       $this->set(ParDataEntity::REVOKE_REASON_FIELD, $reason);
-      // In case a revoke timestamp needs to be applied to an entity date value.
-      $this->setRevokeDateTimestamp();
+
+      // In the case that inspection plans are being revoked as a resulted of a partnership being
+      // revoked we want to prevent historical revoked ISP's from being auto activated in the case
+      // of a partnership being restored.
+      if (!$reason === ParDataPartnership::INSPECTION_PLAN_REVOKE_REASON) {
```
kalpaitch: Just a thought, would a better check be that if the expiry date is already in the past then don't update it?

PR #534, opened 2019-11-25
PAR-1500: Added warning message component (IN PROGRESS)

kalpaitch:
New warning messages added to any Enforcement Notice with at least one approved Enforcement Action:
image

### Wording as follows:
"Please note that this enforcement has been approved and a copy of the information held within must be sent to the business by 26 November 2019 at the latest."

**Note:** The date is 10 working days after the enforcement was approved. Does it need to be stated that this is inclusive, as in it must be sent by the end of 26 November 2019?
**Note:** What do we do with this message _after_ this date has passed? Do we change the message to be in the past tense? Do we hide the message? Do we need to change the wording entirely?


kalpaitch:
In hindsight this needs to be an action for the PA (not the enforcing officer) and thus also needs to be on the EN confirmation page:

image

kalpaitch:
> In hindsight this needs to be an action for the PA (not the enforcing officer) and thus also needs to be on the EN confirmation page:
> 
> image

Apparently this doesn't. The burden is on the enforcement officer to send the notice.

PR #536, opened 2019-11-28
PAR-1502: Updated partnership information.

kalpaitch:
This is the restructuring of the partnership pages _only_. There is a further ticket to turn this page into components so that we don't have to update this information in multiple places but we can get to that later.

Screenshot_2019-11-28 Primary Authority partnership information RECC - Renewable Energy Consumer Code Primary Authority Reg


kalpaitch:
Partnership search page has been amended to look similar, but there are slight differences:

Screenshot_2019-11-29 Partnership Search Primary authority information for RECC - Renewable Energy Consumer Code Primary Au

PR #552, opened 2020-01-30
PAR-1527: Added de-duping of legal entity numbers.

kalpaitch:
@jon-edery What do you think of this, de-duping the legal entities at the stage they're created, I can't forsee any issues as a legal entity will always be returned from `ParDataLegalEntity::create()`, it just might not be a new one.


jon-edery:
this looks good to me I cant see any potential issues here.

PR #555, opened 2020-03-02
PAR-1540 legal entity restrictions

jon-edery:
Screen Shot 2020-03-02 at 17 19 44

### DIFF
--- a/web/modules/custom/par_data/src/Entity/ParDataLegalEntity.php
+++ b/web/modules/custom/par_data/src/Entity/ParDataLegalEntity.php
@@ -180,7 +180,8 @@ class ParDataLegalEntity extends ParDataEntity {
    *  Helper function to determine if a legal entity is referenced by more then one partnership entities.
    *
    * @return Boolean
-   *   TRUE if the legal entity is being referenced in another partnership.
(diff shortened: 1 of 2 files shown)


web/modules/custom/par_data/src/Entity/ParDataLegalEntity.php
```diff
@@ -176,6 +176,18 @@ public function mergeLegalEntities() {
     $this->save();
   }
 
+  /**
+   *  Helper function to determine if a legal entity is referenced by more then one partnership entities.
+   */
+  public function lookupExistingLegalEntities() {
+    $relationships = $this->getRelationships(NULL, NULL, TRUE);
```
kalpaitch: So it's possible to lookup just relationships of a certain type, for example `$this->getRelationships('par_data_partnerships', NULL, TRUE)`.

Also I'm not sure about the name of this method, it should possibly be `isLegalEntityExisting` or `hasExistingPartnerships` and make this more specific to the task at hand??

jon-edery: Good points made the updates suggested.

PR #559, opened 2020-03-06
Circleci

kalpaitch:


scripts/db-backup.sh
```diff
@@ -103,8 +103,12 @@ mkdir -p $DIRECTORY
...
 
 printf "Exporting database dump...\n"
-printf "Running: drush $DRUPAL_ALIAS sql-dump --result-file='$DIRECTORY/$FILE_NAME.sql'\n"
-../vendor/drush/drush/drush $DRUPAL_ALIAS sql-dump --result-file="$DIRECTORY/$FILE_NAME.sql" --verbose
+printf "Running: drush $DRUPAL_ALIAS sql-dump --result-file='$DIRECTORY/$FILE_NAME_raw.sql'\n"
+../vendor/drush/drush/drush $DRUPAL_ALIAS sql-dump --result-file="$DIRECTORY/$FILE_NAME_raw.sql"
+
+## Remove erroneous pg_dump roles applied despite --no-owner flags.
+## https://stackoverflow.com/questions/60605491/using-pg-dump-with-no-owner-information-still-outputting-role-information
+sed -e '/-- do not execute if not member of manager role/,+4d' $DIRECTORY/$FILE_NAME_raw.sql > $DIRECTORY/$FILE_NAME.sql
```
kalpaitch: @jon-edery Take a look at this, it's a bit of a dirty hack of the sql file but I can't see any other way to remove the rds role checks that are included in the dump file.

jon-edery: Yea this is not ideal but can be used for now to resolve the ownership data issue seems like you have spent quite a bit of time on this so I think we can use this for now to get around the issue.  and have a tech debt ticket to allocate time to investigate and see if there is standardised way of solving this issue.

jon-edery: lets merge this and see how we get on..

PR #568, opened 2020-03-26
PAR-1550: Added session security modules (DO NOT MERGE)

kalpaitch:
We need to get permission from the team to merge this PR. To confirm that the following rules are acceptable:
**Regular users:** maximum 1 browser session (if they sign into a new computer, they will be signed out of the old computer), users logged out after 6 hours of inactivity.
**Helpdesk users:** maximum 2 browser sessions, users logged out after 2 hours.


kalpaitch:
Discussion with Team:

- Helpdesk to expire after 4 hours.
- Regular users to expire after 6 hours

I think the browser sessions feature isn't working as well (not redirecting to user login) so perhaps we leave this out for now.

PR #571, opened 2020-04-02
PAR-1510 CSV export

jon-edery:
Screen Shot 2020-04-02 at 20 20 13

Screen Shot 2020-04-02 at 20 20 47


kalpaitch:
Looks pretty good. I can see the download link and am able to download the file.

There is still an issue with the number of results exported. I can see the correct number of entries 2118 in the pager and when I select all entries, but when I click through and it starts processing I get "Prepared 190 of 2108 entities for processing."

I am able to download now, but it does seem to have 10 less items in the list. I couldn't work out your explanation for this.

PR #574, opened 2020-04-09
PAR-1561

jon-edery:
resolve cache invalidation issue on the search path for the partnership controller page.


kalpaitch:
Yup, tested this. To re-create, visit the search partnership page `/partnership/{id}` and the manage partnership page `partnership/{id}/authority-details` and _then_ update the about partnership information.

Without this fix the search partnership page does _not_ get updated without a cache clear. With this fix all changes are reflected immediately. Nice work.

PR #576, opened 2020-04-15
PAR-1572: Removed REQUEST_TIME

kalpaitch:


kalpaitch:
Ok removed all mentions of the REQUEST_TIME constant in all our custom code.

I think this is the first sign of how under-prepared all the contrib modules are for Drupal 9, there are 77 mentions in our contrib modules, none of which have been updated yet even though this constant was deprecated about 4 years ago:
image

web/modules/custom/par_data/src/Entity/ParDataEntity.php
```diff
@@ -72,6 +72,15 @@ public function getDateFormatter() {
     return \Drupal::service('date.formatter');
   }
```
jon-edery, line 73: this can be added to the class:   use Drupal\Component\Datetime\TimeInterface;

web/modules/custom/par_actions/src/ParScheduleManager.php
```diff
@@ -7,6 +7,7 @@
 use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Plugin\DefaultPluginManager;
 use Drupal\Core\Logger\LoggerChannelTrait;
+use Drupal\Component\Datetime\Time;
```
jon-edery, line 11: use Drupal\Component\Datetime\TimeInterface;

kalpaitch: Yup

kalpaitch:
Updated all `use` statements. Also made sure that _all_ class references in comments are fully qualified rather than using the name space of the `use` statement, for example `// @var \Drupal\Component\Datetime\TimeInterface` rather than `// @var TimeInterface`. This is to assist with documentation.

PR #580, opened 2020-04-15
PAR-1573 drupal_set_message

jon-edery:
Screen Shot 2020-04-15 at 14 48 42


web/modules/features/par_partnership_application_flows/src/Form/ParConfirmationReviewForm.php
```diff
@@ -14,6 +14,7 @@
 use Drupal\par_data\ParDataException;
 use Drupal\par_flows\Form\ParBaseForm;
 use Drupal\par_partnership_application_flows\ParFlowAccessTrait;
+use Drupal\Core\Messenger\MessengerInterface;
```
kalpaitch, line 17: You don't really need these lines, it seems to be all pretty well contained in the MessengerTrait. But either way is good :)

jon-edery: Yea noticed that after added a few.

PR #582, opened 2020-04-15
PAR-1573 drupal_set_message()

jon-edery:
Screen Shot 2020-04-15 at 14 48 42


kalpaitch:
As mentioned, create a ticket to list all the contrib modules we're using that contain drupal_set_message, so we can keep an eye on them and make sure they update.

PR #583, opened 2020-04-15
PAR-1577 entity manager()

jon-edery:
Screen Shot 2020-04-15 at 15 26 54


kalpaitch:
Be sure to update https://regulatorydelivery.atlassian.net/browse/PAR-1590 with a list of contrib modules that are currently also using this method, if there are any.

PR #590, opened 2020-04-17
PAR-1582 updating deprecated internal method.

jon-edery:


web/modules/custom/par_forms/src/Plugin/ParForm/ParOrganisationSuggestionForm.php
```diff
@@ -75,7 +75,7 @@ public function getElements($form = [], $cardinality = 1) {
       ];
     }
     else {
-      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->getNextRoute(), $this->getRouteParams());
+      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->progressRoute(), $this->getRouteParams());
```
kalpaitch, line 78: I know it didn't have it before, but I'd explicitly call `progressRoute('next')` for this one too.

kalpaitch:
Hey @jon-edery so sometimes you'll get this error message - 
[`ParFlowException: Could not find the next route`](https://8309-107533871-gh.circle-artifacts.com/0/build/nightwatch/screenshots/5-organisation-user-update-direct-partnership/Business-User---Update-Partnership_ERROR_Apr-17-2020-094631-GMT%2B0000.png).

There are some differences between `getnextRoute` and `progressRoute`, for one you have to be more explicit about what the next route is now. `getNextRoute` tried quite hard to guess what the next route should be if one wasn't specified. But `progressRoute` will just tell you to get lost and be more clear. For example `getNextRoute` used to rewind to the beginning of a journey if there was no next step. There may well be other differences, we'll have to see what they are on a case by case basis.

This error attached seems to be related to the 'cancel' action. Cancel used to just go back a step, but it doesn't do this any more, you have to clearly specify which step in the journey started this 'mini-process' and which step it therefore needs to go back to.

PR #595, opened 2020-04-27
PAR-583 get prev route

jon-edery:


web/modules/custom/par_forms/src/Plugin/ParForm/ParOrganisationSuggestionForm.php
```diff
@@ -62,7 +62,7 @@ public function loadData($cardinality = 1) {
   public function getElements($form = [], $cardinality = 1) {
     // Go back to the previous page if there's no search term.
     if (!$this->getFlowDataHandler()->getFormPermValue('organisation_select_search_query')) {
-      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->getPrevRoute(), $this->getRouteParams());
+      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->progressRoute('prev'), $this->getRouteParams());
       return new RedirectResponse($url);
     }
```
jon-edery, line 67: this code re-factoring redirects to the following:  " "/partnership/apply/organisation/address".
The previous URL for this point in the application is /partnership/apply/organisation/name.

In this instance this seems correct to me as here we start at this point:

Screen Shot 2020-04-28 at 10 22 00

The next step is the suggested organisation:

Screen Shot 2020-04-28 at 10 22 13

If the search query cannot find any suggestions we need to go to step 8 in the flow config (par_partnership_application_flows.partnership_application_organisation_address) which progressRoute generates the link to auto redirect correctly.   redirected form (step 8): 

Screen Shot 2020-04-28 at 10 23 26

jon-edery: this is working how it should.  We also have the option to make an explicit redirect key in the flow config for each journey step where this is applicable for example:

application partnership flow 
step 7 (par_partnership_application_flows.partnership_application_organisation_confirm)
we could add:
redirect:
no_query_result:  8

web/modules/custom/par_forms/src/Plugin/ParForm/ParSelectAuthorityForm.php
```diff
@@ -46,7 +46,7 @@ public function getElements($form = [], $cardinality = 1) {
 
     // If no suggestions were found cancel out of the journey.
     if ($required && count($authorities) <= 0) {
-      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->getPrevRoute('prev'), $this->getRouteParams());
+      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->progressRoute('prev'), $this->getRouteParams());
       return new RedirectResponse($url);
     }
```
jon-edery, line 52: This form plugin is used all over the system.  I have tested it in the context of the partnership application journey and the correct step is loaded after the progressRoute() generates a link and redirects.  The form that is loaded in this instance is /partnership/apply/type which is the correct step we are expecting.


Screen Shot 2020-04-28 at 11 48 18

web/modules/custom/par_forms/src/Plugin/ParForm/ParOrganisationSuggestionForm.php
kalpaitch: I'm not sure that is right. I think you need to do what it was doing before, if a user clicks cancel on the "Are you looking for one of these business?" page they should be going back to "Who are you in partnership with?". They should not be going to the address page.

web/modules/custom/par_forms/src/Plugin/ParForm/ParSelectAuthorityForm.php
kalpaitch: Yea it's not `progressRoute('prev')` the operation 'prev' isn't a valid default.

web/modules/custom/par_forms/src/Plugin/ParForm/ParOrganisationSuggestionForm.php
```diff
@@ -62,7 +62,7 @@ public function loadData($cardinality = 1) {
   public function getElements($form = [], $cardinality = 1) {
     // Go back to the previous page if there's no search term.
     if (!$this->getFlowDataHandler()->getFormPermValue('organisation_select_search_query')) {
-      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->getPrevRoute(), $this->getRouteParams());
+      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->progressRoute(), $this->getRouteParams());
```
kalpaitch, line 65: I think in almost all cases you should swap `getPrevRoute()` for `progressRoute('cancel')`, be specific about the operation.

web/modules/custom/par_forms/src/Plugin/ParForm/ParSelectAuthorityForm.php
```diff
@@ -46,7 +46,7 @@ public function getElements($form = [], $cardinality = 1) {
 
     // If no suggestions were found cancel out of the journey.
     if ($required && count($authorities) <= 0) {
-      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->getPrevRoute('prev'), $this->getRouteParams());
+      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->progressRoute(), $this->getRouteParams());
```
kalpaitch, line 49: Again, be specific about the operation.

web/modules/custom/par_forms/src/Plugin/ParForm/ParSelectOrganisationForm.php
```diff
@@ -46,7 +46,7 @@ public function getElements($form = [], $cardinality = 1) {
 
     // If no suggestions were found cancel out of the journey.
     if ($required && count($organisations) <= 0) {
-      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->getPrevRoute('prev'), $this->getRouteParams());
+      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->progressRoute(), $this->getRouteParams());
```
kalpaitch, line 49: Again, be specific about the operation.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsOrganisationSuggestionForm.php
```diff
@@ -39,7 +39,7 @@ public function buildForm(array $form, FormStateInterface $form_state) {
     // Go to previous step if search query is not specified.
     if (!$searchQuery) {
       // @TODO Find a way to notify the user they have been redirected.
-      return $this->redirect($this->getFlowNegotiator()->getFlow()->getPrevRoute(), $this->getRouteParams());
+      return $this->redirect($this->getFlowNegotiator()->getFlow()->progressRoute(), $this->getRouteParams());
```
kalpaitch, line 42: Again, be specific about the operation.

web/modules/custom/par_flows/tests/src/Unit/ParFormFlowEntityTest.php
```diff
@@ -213,12 +213,12 @@ public function testGetNextRoute() {
    * @covers ::getPrevRoute
    */
   public function testGetPrevRoute() {
-    $prev_route = $this->testFlow->getPrevRoute();
+    $prev_route = $this->testFlow->progressRoute();
```
kalpaitch, line 216: `progressRoute()` isn't the same as `getPrevRoute()`. The former function moves back by default. The new function moves forward by default.

web/modules/custom/par_forms/src/Plugin/ParForm/ParOrganisationSuggestionForm.php
jon-edery: this is my mistake here I have corrected the error. the only thing which is what originally confused me is that if you click continue and the query does not return any results the page just refreshes without telling the user what's going on.  I would suggest a drupal_set_message  the organisation name provided did not match any results..

web/modules/custom/par_flows/tests/src/Unit/ParFormFlowEntityTest.php
```diff
@@ -215,15 +221,29 @@ public function testGetNextRoute() {
...
+    $this->assertEquals('par_test_forms.first', $prev_route, "The previous route has been correctly identified.");
+
+    $prev_route = $this->testFlow->progressRoute('cancel');
+
+    // Check the flow progress correctly for a given operation.
+    $this->assertEquals('par_test_forms.confirmation', $prev_route, "The previous route has been correctly identified given an operation.");
+  }
   /**
    * @covers ::getStepByFormId
    */
```
jon-edery, line 249: Have corrected the error with the unit tests.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsOrganisationSuggestionForm.php
jon-edery: Looks like this form is not being used I cannot find where its being referenced looks like this may have been replaced by and form plugin (ParOrganisationSuggestionForm).   Can you please confirm?

web/modules/custom/par_forms/src/Plugin/ParForm/ParOrganisationSuggestionForm.php
jon-edery: Please note the only journey that loads this form plugin is the partnership application journey.

kalpaitch:
Looks great Jon, nice work.

I've been wondering for a while if the rules that I created in `progressRoute()` are enough, I had two thoughts, mind chatting about these at some point?:
* 'prev_route' seems like a default operation to me and I think we could possibly proscribe it as this to avoid having to put that in the journey config each time, might make sense to call it 'back' (I think the forward one is 'next').
* I think we might as well add in a hook / listener into this method to allow us more flexibility going forward, seems like an easy thing to do now.

PR #597, opened 2020-04-29
PAR-1582 get next route

jon-edery:


web/modules/custom/par_forms/src/Plugin/ParForm/ParSelectPersonForm.php
```diff
@@ -58,13 +58,13 @@ public function getElements($form = [], $cardinality = 1) {
 
     // If there are no people we should create a new record.
     if (count($user_people) <= 0) {
-      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->getNextRoute('cancel'), $this->getRouteParams());
+      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->progressRoute('cancel'), $this->getRouteParams());
       return new RedirectResponse($url);
```
jon-edery, line 62: Looking at the code here it looks like this code -> getNextRoute('cancel'), should have been getNextRoute('next'),
As the next logical step if no duplicates are found is to progress to the next form instead of canceling out of the journey.

web/modules/custom/par_forms/src/Plugin/ParForm/ParDedupePersonForm.php
```diff
@@ -97,13 +97,13 @@ public function getElements($form = [], $cardinality = 1) {
     // If there are no people we should treat it as a new record and skip.
     if (count($contact_records) <= 0) {
       $this->getFlowDataHandler()->setTempDataValue('contact_record', self::ADD_NEW);
-      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->getNextRoute('cancel'), $this->getRouteParams());
+      $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->progressRoute('cancel'), $this->getRouteParams());
       return new RedirectResponse($url);
     }
```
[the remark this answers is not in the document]
kalpaitch: This form appears to be used in one place only at the moment: `partnership_contact_add`

It would appear you're right that the step in the journey that represents this page does not have a 'cancel' operation:
```

web/modules/custom/par_forms/src/Plugin/ParForm/ParSelectPersonForm.php
kalpaitch: Agreed as above.

kalpaitch:
I've realised something else that may make your job more fun. There's another couple of functions `getNextLink` and `getPrevLink`

I've searched the codebase and `getPrevLink` is not being used, please confirm and then @deprecate and remove it in the same way.

`getNextLink` is used in a number of places, luckily it seems none of these use any default operations such as 'cancel'. We will need to think what to do about this, I suspect we just leave it as is, or maybe rename. The internal mechanisms of this now call `progressRoute` anyway so I think this is just a naming issue.

PR #598, opened 2020-04-30
PAR-1504: Added initial partnership page components

kalpaitch:
The purpose of this feature is to turn all the current components of the partnership page into 'form components' so that they can be re-used on the three pages where they appear:
- Authority partnership management page
- Organisation partnership management page
- General partnership page (search partnership journey) *

* note that these plugins haven't yet been applied to the search page, I think a better first step is to apply them to the other pages first.

We can consider using these component blocks for the partnership application review stages also. But will leave that pending for another ticket.

There have been a number of bugs where one of these pages has displayed different information to the other, the hope is that **this enables us to be more consistent with the display of partnerships**.

It should also allow us customise these components more freely. They were previously all generated by the `ParDisplayTrait` which meant that all the fields had to follow a similar display, including the number and position of links and the link text itself. **This update should allow us to add links more freely**, for example we can now add custom links, including those required to delete or remove [from the partnership only] data.

## Scope
The scope of this ticket was just to turn the current page into components, this has for the most part left most of the elements looking and working the same. There are a few exceptions, however:
- some of the links have been lowercased to be more consistent
- I've tried out putting the Advice & Inspection plan links in columns, to break up the display and save space. It's very minor and probably not going to matter much, it can be easily reversed if not liked.
- resolving a few bugs while going through this, including the display of members and the number of members. Which includes removing the legacy 'business size' field which talked about 'small, medium and large' businesses and is no longer required, changing the display of members field so that it is clear that we're counting 'active' members only, and making sure this field does only count active members.
- resolving the bug whereby members other than the organisation can confirm the terms and conditions.

The other big change from a technical standpoint is that I've had to allow plugins to be used multiple times on a page (for example the contact display for authorities and then organisations). This has meant that the yaml key in the flow configuration under 'components' is no longer the _default_ `plugin_name`, it is possible to set this key to an entirely random value and set the `plugin_name` configuration parameter on this component instead. I think this works, is backward compatible and makes some sense.

I've also realised in doing this that really each of these components should be plugin derivatives rather than the same plugin instances, but for now this works so I'm not going to go there :)


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsDetailsForm.php
```diff
@@ -156,7 +71,7 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
 
     // View and perform operations on the information about the business.
     $form['about_business'] = $this->renderSection('About the organisation', $par_data_organisation, ['comments' => 'about'], ['edit-field']);
-
+var_dump($form['registered_address']); die;
```
jon-edery, line 74: debug

kalpaitch: Yea, still got a lot of work to do on this, just putting the PR for my own benefit atm :)

sync/par_flows.par_flow.partnership_direct.yml
```diff
@@ -16,12 +16,35 @@ steps:
...
+        display: organisation
+        show_title: false
+        weight: -2
     redirect:
       edit_comments: 2
       add_field_premises: 3
       edit_field_premises: 4
       add_field_legal_entity: 5
       edit_field_legal_entity: 6
+      remove_field_legal_entity: 6
```
jon-edery, line 47: Should this operation be set within the config for all journey's?

jon-edery: All the partnership journeys are using this plugin "id = "legal_entity_display" which contains login which generates a link bt operation -> >getLinkByCurrentOperation('remove_field_legal_entity', $params, $options, TRUE);

the operation remove_field_legal_entity is only set within one out of the 3 journeys.

web/modules/features/par_partnership_flows/config/install/par_flows.par_flow.partnership_authority.yml
```diff
@@ -13,6 +13,30 @@ steps:
   1:
     route: par_partnership_flows.authority_details
     form_id: par_partnership_details
+    components:
+      partnership_information_display:
+        weight: -10
+      organisation_information_display:
+        weight: -9
+      partnership_members:
+        weight: -8
```
jon-edery, line 22: should the formate be applied here " format: member_link_view" ?

sync/par_flows.par_flow.partnership_direct.yml
kalpaitch: Yea, so on the first point, at first I thought that this link should only be on the direct partnership journey but no you mention it I'm not so sure why. I haven't actually added this anyway, that was going to be in another ticket so in terms of this PR it should probably be removed.

On the second point, if a plugin contains a `getLinkByCurrentOperation('remove_field_legal_entity', ...)` and the operation doesn't exist in the journey _or_ isn't accessible by the current user, then the link will simply not be displayed (there's a try {} catch) which in this instance is the expected behaviour since this particular link should only show in very specific circumstances.

web/modules/features/par_partnership_flows/config/install/par_flows.par_flow.partnership_authority.yml
kalpaitch: The default view is `format: member_list` so doesn't really need to be stated in this instance.

PR #601, opened 2020-05-12
PAR-1569: Added link to remove a legal entity from the partnership

kalpaitch:
Added a link to remove the legal entity from the partnership. This can only be used if the following conditions are _all_ met:
* There are more than one legal entity on the partnership
* The partnership is _not_ active
* The user is a helpdesk user

### DIFF
--- a/web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsRemoveLegalEntityForm.php
+++ b/web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsRemoveLegalEntityForm.php
@@ -54,7 +54,7 @@ class ParPartnershipFlowsRemoveLegalEntityForm extends ParBaseForm {
     // If there is no legal entity skip this step.
     // @TODO Monitor PAR-1592. If a PR is submitted for that it will need
     // applying jto this method call too.
-    if (null === $delta) {
+    if ($delta === NULL) {
       $url = $this->getUrlGenerator()->generateFromRoute($this->getFlowNegotiator()->getFlow()->progressRoute('cancel'), $this->getRouteParams());
       return new RedirectResponse($url);
     }
--- a/web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsRemoveLegalEntityForm.php
+++ b/web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsRemoveLegalEntityForm.php
@@ -3,81 +3,82 @@
 namespace Drupal\par_partnership_flows\Form;
 
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\par_data\Entity\ParDataEntity;
+use Drupal\par_data\Entity\ParDataEntityInterface;
 use Drupal\par_data\Entity\ParDataPartnership;
 use Drupal\par_flows\Form\ParBaseForm;
 use Drupal\par_partnership_flows\ParPartnershipFlowAccessTrait;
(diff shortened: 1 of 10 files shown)


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsRemoveLegalEntityForm.php
```diff
@@ -0,0 +1,127 @@
...
+  public function buildForm(array $form, FormStateInterface $form_state, ParDataPartnership $par_data_partnership = NULL, $field_legal_entity_delta = NULL) {
+    $par_data_partnership = $this->getFlowDataHandler()->getParameter('par_data_partnership');
+    $par_data_legal_entity = $this->getFlowDataHandler()->getParameter('par_data_legal_entity');
+
+    $delta = $this->getFlowDataHandler()->getFormPermValue('field_legal_entity_delta');
+
+    // If there is no legal entity skip this step.
+    // @TODO Monitor PAR-1592. If a PR is submitted for that it will need
+    // applying jto this method call too.
+    if (null === $delta) {
```
jon-edery, line 57: $delta === NULL

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsRemoveLegalEntityForm.php
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * The partnership form for removing the legal entity.
+ */
+class ParPartnershipFlowsRemoveLegalEntityForm extends ParBaseForm {
+
+  use ParPartnershipFlowsTrait;
+  use ParPartnershipFlowAccessTrait;
+
+  protected $pageTitle = 'Are you sure you want to remove the legal entity?';
```
jon-edery, line 22: to remove the legal entity?  =>  to remove this legal entity?

web/modules/features/par_partnership_flows/src/ParPartnershipFlowAccessTrait.php
```diff
@@ -75,6 +75,13 @@ public function accessCallback(Route $route, RouteMatchInterface $route_match, A
 
         break;
 
+
+      case 'par_partnership_flows.legal_entity_remove':
+        // Prohibit removing of the last item.
+        if ($par_data_partnership->get('field_legal_entity')->count() <= 1) {
```
jon-edery, line 81: condition needed for active partnerships ->  // Restrict advice upload to active partnerships only.
        if (!$par_data_partnership->isActive()) {

kalpaitch: So because there is no `break;` clause in this case it will continue on and do the next steps also... which includes the `!$par_data_partnership->isActive()` check.

If this isn't clear we can update, but probably the easiest option for now.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsRemoveLegalEntityForm.php
kalpaitch: Much of a muchness, but I think you're right, I think that is how we've done it everywhere else.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsRemoveLegalEntityForm.php
kalpaitch: Nice.

PR #603, opened 2020-05-13
PAR-1569

kalpaitch:
Added a link to remove the legal entity from the partnership. This can only be used if the following conditions are _all_ met:
* There are more than one legal entity on the partnership
* The partnership is _not_ active
* The user is a helpdesk user

See PR to parent branch (now merged) for original comments - https://github.com/UKGovernmentBEIS/beis-primary-authority-register/pull/601

### DIFF
--- a/web/modules/features/par_partnership_flows/src/ParPartnershipFlowAccessTrait.php
+++ b/web/modules/features/par_partnership_flows/src/ParPartnershipFlowAccessTrait.php
@@ -75,6 +75,13 @@ trait ParPartnershipFlowAccessTrait {
 
         break;
 
+
+      case 'par_partnership_flows.legal_entity_remove':
+        // Prohibit removing of the last item.
+        if ($par_data_partnership->get('field_legal_entity')->count() <= 1) {
+          $this->accessResult = AccessResult::forbidden('The last legal entity can\'t be removed.');
+        }
+
       case 'par_partnership_flows.legal_entity_add':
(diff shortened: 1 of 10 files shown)


web/modules/features/par_partnership_flows/src/ParPartnershipFlowAccessTrait.php
```diff
@@ -75,6 +75,13 @@ public function accessCallback(Route $route, RouteMatchInterface $route_match, A
 
         break;
 
+
+      case 'par_partnership_flows.legal_entity_remove':
+        // Prohibit removing of the last item.
+        if ($par_data_partnership->get('field_legal_entity')->count() <= 1) {
+          $this->accessResult = AccessResult::forbidden('The last legal entity can\'t be removed.');
+        }
+
```
jon-edery, line 84: I can see there is no break statement here but the route will still have to match one of the cases below " case 'par_partnership_flows.legal_entity_add':  case 'par_partnership_flows.legal_entity_edit':"

in order for the " if ($par_data_partnership->isActive())  " check to be executed.  

unless I am missing something here I would recommend to the following:

  //  Evaluate access conditions for removing legal entities.
        if ($par_data_partnership->get('field_legal_entity')->count() <= 1 || $par_data_partnership->isActive()) {
          $this->accessResult = AccessResult::forbidden('This last legal entity can\'t be removed.');
        }

PR #605, opened 2020-05-19
PAR-1582 get next route listener

jon-edery:
Original PR ->  https://github.com/UKGovernmentBEIS/beis-primary-authority-register/pull/600

So, like I was saying yesterday I found an instance where a form plugin automatically cancels the journey if there were no results (authorities selected i.e. a helpdesk user): authority_select form plugin.

Because we've changed the default behaviour of progressRoute() this now doesn't fallback on anything. Anyway we talked about this and we talked about having a hook/alter/listener to allow us to add fallbacks and configure them...

... so I went looking around and found that we already have a ParFlowEvent that we dispatch in: ParBaseForm::submitForm(), which fires when any button other than cancel is used to submit a form; ParBaseForm::cancelForm(), which fires when the cancel button of a form is submitted; ParBaseController::getProceedingUrl(), which is a helper function used to generate the links in all static controllers.

Anyway, it turns out that each of these three methods that call progressRoute() also call a couple of other default fallbacks. This isn't the same in each one though, we'll have to look into this, and it only happens for the main form buttons (at the bottom of the form or page).

I've done a proof on concept to give a bit of an idea of what I'm on about, basically my current thinking is that the event dispatcher shoud do so from the progressRoute() method so that it fires (and can look for fallbacks) every time the method is called. I also feel that the couple of 'other defaults' that are in each of the three methods above should if possible be turned into event subscribers which would allow us to configure these default behaviours to run on the conditions we choose.

Needs a bit more thought, and a separate ticket.

One last thing, I noticed that there were two other methods ParFlow::getNextLink() and ParFlow::getPrevLink(), we don't need to convert them per se, but we will need to make them go through progressRoute, so that all of the 'next step' decisions are being made in one place (currently they're calling ParFlow::getNextStep() which runs different progressing logic again).


web/modules/custom/par_flows/src/Event/ParFlowEvents.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace Drupal\par_flows\Event;
+
+/**
+ * Defines events for the par_flows module.
```
kalpaitch, line 6: Agreed

web/modules/custom/par_flows/src/Event/ParFlowEvent.php
```diff
@@ -54,14 +60,37 @@ class ParFlowEvent extends Event {
...
+
+    if ($entryUrl) {
+      $this->entryUrl = $entryUrl;
+    }
+  }
+
+  /**
+   * Static method for generating an event name based on the operation.
+   */
+  public static function getCustomEvent($operation) {
```
kalpaitch, line 80: I think this method should probable move to the ParFlowEvents class then if we're separating the even naming class.

web/modules/custom/par_flows/src/EventSubscriber/ParFlowCancelSubscriber.php
```diff
@@ -0,0 +1,62 @@
...
+  public function onEvent(ParFlowEventInterface $event) {
+
+    $redirect_url = $event->getUrl();
+
+    // If the redirect URL has already been set by the previous conditions this event subscriber will not set a fallback default.
+    if (isset($redirect_url) && ($redirect_url instanceof Url)) {
+      return;
+    }
+
+    // If the cancel route could be found in the flow then
```
kalpaitch, line 42: I think it would be good to have separate event subscribers for each of the 'rules' that we we can easily apply them to different flows:
- EntryUrl
- Dashboard fallback (needs to switch between helpdesk and normal user dashboard using context)

There may be others that we need to apply for certain journeys. We can find the most appropriate way to configure these.

kalpaitch: Also in terms of the event subscribers they don't need to be specific to Cancelling or Submitting, for example the two default rules we have here 'EntryUrl' and 'DashboardFallback' occur on both submit and cancel, we can just apply two events to the same listener:
```
  static function getSubscribedEvents() {
    $events[ParFlowEvent::getCustomEvent('next')][] = ['onEvent', -100];
    $events[ParFlowEvent::getCustomEvent('cancel')][] = ['onEvent', -100];
    return $events;
  }
```

web/modules/custom/par_flows/src/EventSubscriber/ParFlowSubmitSubscriber.php
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * The events to react to.
+   *
+   * @return mixed
+   */
+  static function getSubscribedEvents() {
+
+    // Listen for forms being submitted with the "next" $submit_action.
+    $events[ParFlowEvent::getCustomEvent('next')][] = ['onEvent', -100];
```
kalpaitch, line 25: Also we have a bit of a problem with the naming of events, we'll need to act on every known submission action because the action is in the event name.

I don't know how we get around this, especially when there could be an unknown number of actions, we can't really catch them all???

kalpaitch: I think this was legacy naming, if we want to change the event name we'd need to make sure it didn't impact any existing functionality.

web/modules/custom/par_flows/src/Entity/ParFlow.php
```diff
@@ -339,7 +362,8 @@ public function getPrevStep($operation = NULL) {
   /**
    * {@inheritdoc}
    */
-  public function progressRoute($operation = NULL) {
+  public function progressRoute($operation = NULL, $entry_point_URL = NULL) {
```
kalpaitch, line 365: IMO `$entry_point_URL` should not be defined here, this is generic functionality that shouldn't need to know anything about the state of the flow. We need to be able to access this data from `$this->getFlowDataHandler()->getMetaDataValue(ParFlowDataHandler::ENTRY_POINT)' instead like all other cases.

web/modules/custom/par_flows/src/Entity/ParFlow.php
```diff
@@ -348,31 +372,40 @@ public function progressRoute($operation = NULL) {
...
+      $redirect_route = $redirect_step['route'] ?? NULL;
     }
 
-    // Rule 4) Allow other modules to alter the redirection rules.
-    // @TODO Add a hook alter to allow additional redirection rules per module.
+    $redirect_url = isset($redirect_route) ? Url::fromRoute($redirect_route, $this->getRouteParams()) : NULL;
 
-    if (empty($redirect_step)) {
+    // Rule 4) Allow other modules to alter the redirection rules.
+    $event = new ParFlowEvent($this, $this->getCurrentRouteMatch(), $redirect_url, $entry_point_URL);
```
kalpaitch, line 397: If at all possible I think it would be good to separate specific data such as the `$entry_point_URL` from the call for this event. If we can pass the state of the flow in such that any event subscriber can extract whatever data it needs that'd be ace.

web/modules/custom/par_flows/src/EventSubscriber/ParFlowCancelSubscriber.php
jon-edery: good point.

PR #612, opened 2020-06-09
PAR-1619: Adding tests for entity merging.

kalpaitch:
@jon-edery I'm adding some tests for entity merging and notice a few errors with this, noticeably it fails if the contact we're trying to merge doesn't have a user account associated with it.

The noteable changes are:
* Moved the `merge()` method to the base ParDataEntity so that all entities of the same type can be merged.
* Added unit tests to confirm this is working as expected.
* Removed the merge option from the typical update person journey in an effort to simplify this journey
* Added a separate journey just for merging


jon-edery:
I might be worth trying this module out looks like the last release was on 06/07/20 and the guy has added Drupal 9 compatibility.  but very much depends on how stable what we got is we could always use some on the code in this module to re factor what we currently have.

kalpaitch:
> I might be worth trying this module out looks like the last release was on 06/07/20 and the guy has added Drupal 9 compatibility. but very much depends on how stable what we got is we could always use some on the code in this module to re factor what we currently have.


I'm planning to change all this, but wondering if we should consider using this contrib module - https://git.drupalcode.org/project/simple_entity_merge/-/blob/8.x-1.x/src/SimpleEntityMerge.php

PR #613, opened 2020-06-11
PAR-1618

jon-edery:
Screen Shot 2020-06-12 at 20 07 57
Screen Shot 2020-06-12 at 20 07 27
Screen Shot 2020-06-12 at 20 06 43
Screen Shot 2020-06-12 at 20 05 16
Screen Shot 2020-06-12 at 20 04 57


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAboutForm.php
```diff
@@ -46,7 +46,12 @@ public function titleCallback() {
    */
   public function retrieveEditableValues(ParDataPartnership $par_data_partnership = NULL) {
     if ($par_data_partnership) {
-      $this->getFlowDataHandler()->setFormPermValue('about_partnership', $par_data_partnership->get('about_partnership')->getString());
+      // PAR-1618 ensure that partnerships which have been set with the basic_html formatter via the admin
+      // don't output HTML tags on edit forms. This is resulting in the tags being saved with content updates when users
+      // edit the about text.
+      $value = strip_tags($par_data_partnership->get('about_partnership')->getString());
```
kalpaitch, line 52: There are Drupal functions for cleaning the data.

Also This is probably what you want `$par_data_partnership->about_partnership->view('full')`

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAboutForm.php
```diff
@@ -46,7 +46,12 @@ public function titleCallback() {
    */
   public function retrieveEditableValues(ParDataPartnership $par_data_partnership = NULL) {
     if ($par_data_partnership) {
-      $this->getFlowDataHandler()->setFormPermValue('about_partnership', $par_data_partnership->get('about_partnership')->getString());
```
kalpaitch, line 49: Also I suspect there is the same issue for the about organisation text.

web/modules/custom/par_forms/src/Plugin/ParForm/ParAboutPartnershipForm.php
```diff
@@ -29,7 +29,7 @@ class ParAboutPartnershipForm extends ParFormPluginBase {
    * {@inheritdoc}
    */
   public function loadData($cardinality = 1) {
-    if ($par_data_partnership = $this->getFlowDataHandler()->getParameter('par_data_organisation')) {
+    if ($par_data_partnership = $this->getFlowDataHandler()->getParameter('par_data_partnership')) {
       $this->getFlowDataHandler()->setFormPermValue('about_partnership', $par_data_partnership->get('about_partnership')->getString());
     }
```
jon-edery, line 35: Looks like this plugin is not in use?

kalpaitch: Ha.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAboutForm.php
kalpaitch: I've also been looking for the actual field widget that displays long text, I think it might be:
```
$form['body'] = array(
    '#type' => 'text_format',
    '#title' => 'Body',
    '#format' => 'full_html',
    '#default_value' => '<p>The quick brown fox jumped over the lazy dog.</p>',
);
```

Probably worth having a little look at this to see if this gives us what we want by default without having to reinvent our own field. It might be that this requires a text format selector to be displayed though.

jon-edery: Rendering the element with text_format "full_html" will not removed the ", basic_html" text which has been saved as part of the plain text for the about field seems to be due to some historical error with this value being set through the admin.   there are around 900 + partnerships that have this custom format name stored as part of the about text.

kalpaitch: Ah. crap yea.

PR #615, opened 2020-06-17
PAR-1618  error format name stored in some partnerships 

jon-edery:


kalpaitch:
Nice :)

Does this happen with any of the other fields?

web/modules/custom/par_data/par_data.install
```diff
@@ -999,3 +999,20 @@ function par_data_update_8042() {
...
+
+  // Get all partnerships that have the format issue within the about field.
+  $query = \Drupal::entityQuery('par_data_partnership');
+  $query->condition('about_partnership__value', ', basic_html', 'CONTAINS');
+  $entity_ids = $query->execute();
+  $partnerships_to_update = \Drupal::entityTypeManager()->getStorage('par_data_partnership')->loadMultiple($entity_ids);
+
+  foreach ($partnerships_to_update as $par_data_partnership) {
+    // Strip out the custom format embedded in the text value.
+    $par_data_partnership->about_partnership->value = str_replace(', basic_html', '', $par_data_partnership->about_partnership->value);
```
kalpaitch, line 1015: Just noticed it might be worth checking the `, full_html` and `, restricted_html` filters haven't been used anywhere too.

kalpaitch: On second thoughts I can't find any in my database, it's probably fine.

jon-edery: yea same I have looked at recent version of the database

jon-edery:
> Does this happen with any of the other fields?

this specific issue I can only find this on the about partnership field and only on 32 instances.

PR #617, opened 2020-06-19
PAR-1618 render markup updates

jon-edery:
Screen Shot 2020-06-22 at 10 31 33
Screen Shot 2020-06-22 at 10 12 43


kalpaitch:
Mind giving us a screenshot of what that looks like?

jon-edery:
the screen shots are of both the organisation member and the authority member users editing partnerships. tested the rendering it resolves the embedded markup displaying

kalpaitch:
I thought you weren't going to use this text_format field? It displays the text format information which we don't really want and the field description which we also don't really want.

kalpaitch:
I think the text_format isn't really what we want because of the extra text format box.

I think I questioned whether we want the user to be able to enter html (links especially) because at the moment we can't.

We could create our own render format that is similar to text_format but without the text format box. Or I guess we could just stick with the plain text area that we're using at the moment.

jon-edery:
I have tested both using text_format and this solution with all roles on the different partnership pages.  I think this seems like the cleanest solution as there are a few potential bugs with CK editor as well as a hack needed to disable the additional widget for the about formats info section. I have removed the filter "Display any HTML as plain text" from the plain text formatter as we had the "Convert line breaks into HTML (i.e. <br> and <p>)" which was running first and the any spaces line break HTML produced in the editor being printed out as text which always resulted in a </p> tag being printed on the controller pages. See 

Screen Shot 2020-06-29 at 12 15 20

web/modules/custom/par_flows/src/ParFlowDataHandler.php
```diff
@@ -351,13 +351,17 @@ public function getFormPermValue($key, ParFormPluginInterface $plugin = NULL) {
...
-  public function setFormPermValue($key, $value, ParFormPluginInterface $plugin = NULL) {
+  public function setFormPermValue($key, $value, ParFormPluginInterface $plugin = NULL, $format = 'basic_html') {
     $key = (array) $key;
     // Allow the plugin namespace to be used as a prefix if a plugin is passed in.
     if ($plugin && $plugin instanceof ParFormPluginInterface) {
       array_unshift($key, $plugin->getPluginNamespace());
     }
 
+    if (is_string($value)) {
+      $value = check_markup($value, $format);
```
kalpaitch, line 362: I think we should be following the Drupal way of filtering on output not on input.

jon-edery: agree with that I have changed this now.

jon-edery: This method does not store store modified text values to the database the setFormPermValue saves the processed text to the local cache before if is output as default values to the form so essentially this is applying the check markup on output.  There is not permanent saving of modified data, the original data is not modified which is inline with Drupal's "filter on output" philosophy.

kalpaitch: Oh yea I see. In regards to the first point this will still apply check_markup to all strings (which may even include typecast boolean or integer fields), shouldn't we apply this specifically to text_long fields?

sync/filter.format.plain_text.yml
```diff
@@ -8,12 +8,6 @@ name: 'Plain text'
 format: plain_text
 weight: 10
 filters:
-  filter_html_escape:
```
kalpaitch, line 11: I remember you stripped this from here because it was duplicated, just noticed that this is removed from the plain_text filter rather than the par_html filter? Only saying because the plain_html is a standard drupal filter, possibly used in other modules such as email body text...

kalpaitch:
@jon-edery I think I've identified that there is no way to parse html into plain text in Drupal text areas.

Note that if you have a typed some text into a text field with filter par_html you cannot change this to plain_text, the plain_text filter option will have simply disappeared. The only way to convert it back to plain text is to copy the contents of the ckeditor, delete it, save the entity, change the filter back to plain_text while the field is empty, and then paste in.

This means that any fields that have already been setup as par_html (those that were edited through the admin ui) cannot be converted back...

I did find a work around which is to use the MailFormatter which does parsing of basic html to text - https://github.com/UKGovernmentBEIS/beis-primary-authority-register/pull/617/commits/40b5e89aa611e325e10c76aa1d76ca4fc2b61596

This seems to resolve the issue, in combination with making sure that if we ever save something in the admin ui in the future we've got to make sure we select the plain_text filter before we enter any content in the field.

PR #618, opened 2020-06-22
PAR-1618 clean org text

jon-edery:
found the same issue as the about partnership text. on organisations (see image).

Screen Shot 2020-06-22 at 11 42 38


jon-edery:
image taken from production.

jon-edery:
Testing results before applied update (production database):

Screen Shot 2020-06-23 at 22 22 39

After update has been applied:

Screen Shot 2020-06-23 at 22 24 21

PR #621, opened 2020-06-25
PAR-1592 url generator

jon-edery:
* Dynamically get url generator service.
   * Override the standard way the service is being loaded via UrlGeneratorTrait
   * As this trait is deprecated and will be removed in Drupal 9.


jon-edery:
Not that form plugins are covered via the getUrlGenerator() method in the ParFormPluginBase class

web/modules/custom/par_flows/src/Form/ParBaseForm.php
```diff
@@ -143,6 +143,18 @@ public function getCacheContexts() {
     return ['user.roles', 'route'];
   }
 
+  /**
```
kalpaitch, line 146: It's better to load this in once when the base form is instantiated rather than call it again for every time a plugin calls this method.

jon-edery: yea good point ill update this

web/modules/custom/par_flows/src/ParControllerTrait.php
```diff
@@ -170,6 +178,13 @@ public function getFormBuilder() {
     return $this->formBuilder;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function getUrlGenerator() {
```
kalpaitch, line 184: If this method in in the trait then you probably need to inject the url_generator  into the ParBaseController too, that uses this trait as well.

kalpaitch:
I believe `use Drupal\Core\Routing\UrlGeneratorInterface;` needs to be defined for all these classes.

PR #625, opened 2020-06-30
PAR-1618: Updated plain text filtering.

kalpaitch:
There are three things that need to happen simultaneously to make sure we don't get filter inputs in the text area:

1. Save the format `plain_text` every time.
2. Don't call `getString()` to but `get('value')->getValue()` and extract the raw value instead of the combined value.
3. Sanitise all existing cases where the filter input is already in the field value as per the hook_8044

If we don't do 1) then all the current values that have been saved _without_ a format will 'assume' a format when next that page is edited in the CKeditor, and usually they'll assume the wrong format (hence the missing formatting).

If we don't do 2) then all the current values that have been saved _with_ a format will output string+format when next they are edited in any of the journeys... and they will therefore be saved with this format appended to the value in the value field.

### DIFF
--- a/web/modules/custom/par_data/src/ParDataStorage.php
+++ b/web/modules/custom/par_data/src/ParDataStorage.php
@@ -117,6 +117,18 @@ class ParDataStorage extends TranceStorage {
       }
     }
 
+    // Ensure that a format value is selected for all text_long field values.
+    // @SEE PAR-1618: Text formats not being correctly set or retrieved.
+    foreach ($entity->getFieldDefinitions() as $definition) {
+      if ($definition->getType() === 'text_long' && !$entity->get($definition->getName())->isEmpty()) {
+        foreach ($entity->get($definition->getName()) as $delta => $value) {
+          if ($value->get('format')->getValue() === NULL) {
+            $entity->get($definition->getName())->get($delta)->get('format')->setValue('plain_text');
+          }
+        }
+      }
+    }
+
     // Load the original entity if it already exists.
     if ($this->has($entity->id(), $entity) && !isset($entity->original)) {
       $entity->original = $this->loadUnchanged($entity->id());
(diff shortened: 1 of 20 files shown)


web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAboutBusinessForm.php
```diff
@@ -77,7 +73,10 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
     // Save the value for the about_partnership field.
     $partnership = $this->getFlowDataHandler()->getParameter('par_data_partnership');
     $par_data_organisation = current($partnership->getOrganisation());
-    $par_data_organisation->set('comments', $this->getFlowDataHandler()->getTempDataValue('about_business'));
+    $par_data_organisation->get('comments')->setValue([
+      'value' => $this->getFlowDataHandler()->getTempDataValue('about_business'),
+      'format' => 'plain_text',
```
kalpaitch, line 78: The important thing is that we always save the value with the format so that it renders correctly in the admin interface.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsAboutForm.php
```diff
@@ -46,8 +46,7 @@ public function titleCallback() {
    */
   public function retrieveEditableValues(ParDataPartnership $par_data_partnership = NULL) {
     if ($par_data_partnership) {
-      $information_display = $par_data_partnership->about_partnership->value;
-      $this->getFlowDataHandler()->setFormPermValue('about_partnership', $information_display, NULL, TRUE);
+      $this->getFlowDataHandler()->setFormPermValue('about_partnership', $par_data_partnership->getPlain('about_partnership'));
```
kalpaitch, line 49: And that we get the plain value only from the field rather than the combined string+format value.

web/modules/custom/par_data/src/ParDataStorage.php
```diff
@@ -117,6 +117,18 @@ public function save(EntityInterface $entity) {
       }
     }
 
+    // Ensure that a format value is selected for all text_long field values.
+    // @SEE PAR-1618: Text formats not being correctly set or retrieved.
```
kalpaitch, line 121: This is just a backup so that we don't have to go looking for all the places where a text_long field is saved.

We could potentially remove this once we've identified all those places and saved them with the format.

PR #635, opened 2020-07-15
PAR-1641

jon-edery:
fix the regex pattern to validate on the custom Gds date element.

Error:

Screen Shot 2020-07-15 at 17 01 51

### DIFF
--- a/web/modules/custom/par_forms/src/Element/GdsDate.php
+++ b/web/modules/custom/par_forms/src/Element/GdsDate.php
@@ -62,7 +62,7 @@ class GdsDate extends FormElement {
       '#title' => 'Day',
       '#attributes' => [
         'name' => $element['#name'] ."_day",
-        'pattern' => "([1-9]|[12][0-9]|3[01])",
+        'pattern' => "(0[1-9]|[12]\d|3[01])",
         'size' => 6,
         'class' => ['gds-date-sub-element']
       ],
--- a/web/modules/custom/par_forms/src/Element/GdsDate.php
+++ b/web/modules/custom/par_forms/src/Element/GdsDate.php
@@ -62,7 +62,7 @@ class GdsDate extends FormElement {
       '#title' => 'Day',
       '#attributes' => [
         'name' => $element['#name'] ."_day",
-        'pattern' => "(0[1-9]|[12]\d|3[01])",
+        'pattern' => "(0[1-9]|[12][1-9]|3[01])",
         'size' => 6,
         'class' => ['gds-date-sub-element']
       ],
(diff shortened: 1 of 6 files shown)


web/modules/custom/par_forms/src/Element/GdsDate.php
```diff
@@ -62,7 +62,7 @@ public static function processGdsDate(&$element, FormStateInterface $form_state,
       '#title' => 'Day',
       '#attributes' => [
         'name' => $element['#name'] ."_day",
-        'pattern' => "([1-9]|[12][0-9]|3[01])",
+        'pattern' => "(0[1-9]|[12]\d|3[01])",
```
kalpaitch, line 65: I see, nearly there, make sure the first `0?` is optional, and you best keep the middle statement the same `[12][0-9]` rather than `\d`:
`(0?[1-9]|[12][0-9]|3[01])`

jon-edery: \d === [0-9] just code preference  
0 is optional in the expression ? is not needed.

jon-edery: we could update this regex to conform to the format also 

1|2|3|4|5|6|7|8|9|10|11|12|01|02|03|04|05|06|07|08|09) => 01|02|03|04|05|06|07|08|09|10|11|12)

or just leave it as is ...

web/modules/custom/par_forms/src/Element/GdsDate.php
```diff
@@ -75,7 +75,7 @@ public static function processGdsDate(&$element, FormStateInterface $form_state,
       '#title' => 'Month',
       '#attributes' => [
         'name' => $element['#name'] ."_month",
-        'pattern' => "(1|2|3|4|5|6|7|8|9|10|11|12|01|02|03|04|05|06|07|08|09)",
+        'pattern' => "(01|02|03|04|05|06|07|08|09|10|11|12)",
```
kalpaitch, line 78: We don't want to force the format to be 01-01-2020, we want to allow both that and 1-1-2020 to be permissible no??

kalpaitch:
Also let's put some tests in to check all these formats work.

web/modules/custom/par_forms/src/Element/GdsDate.php
jon-edery: Not sure this is a standard date format 8/1/2020 or 01/2/2020 in forms (does not match the format we asks user to conform to).  

I think its better to allow a single consistent date format for date entry.

jon-edery: by restricting the number of ways the same data can be submitted to the system we reduce the risk of unforeseen bugs.  what do u think?

jon-edery:
yea agree would be good to add some tests

web/modules/custom/par_forms/src/Element/GdsDate.php
kalpaitch: It's a legitimate date format - https://design-system.service.gov.uk/components/date-input/

The GDS guidelines are _do not restrict_ legitimate formats. I thought we were just re-formatting one format into another, we shouldn't be disallowing one format.

PR #638, opened 2020-07-21
PAR-1639 

jon-edery:
add corrected form validation rules to prevent active advice not being listed error.


web/modules/custom/par_forms/src/ParEntityValidationMappingTrait.php
```diff
@@ -140,6 +140,9 @@ public function buildEntity(&$entity, $values) {
         continue;
       }
       $value = NestedArray::getValue($values, (array) $mapping->getElement());
+      // Cleanse the value just a little bit.
+      $value = is_array($value) ? array_filter($value) : $value;
```
kalpaitch, line 144: Hey @jon-edery turns out that array filter you were including in the submit function was needed in the mapping validation too.

I think this basic check is safe enough, I can't imagine any situation where a form submits an array with empty values that we want to preserve...

PR #641, opened 2020-07-27
PAR-1633 adding  phone number format validation

jon-edery:
Enforce basic accepted phone number validation.  
prevent string on tel fields.

This format validation does not enforce the field to be required validation is only enforce on data entry.

### DIFF
--- a/web/modules/custom/par_forms/src/Plugin/ParForm/ParContactDetailsBasicForm.php
+++ b/web/modules/custom/par_forms/src/Plugin/ParForm/ParContactDetailsBasicForm.php
@@ -63,6 +63,9 @@ class ParContactDetailsBasicForm extends ParFormPluginBase {
       '#type' => 'tel',
       '#title' => $this->t('Enter the work phone number'),
       '#default_value' => $this->getDefaultValuesByKey('work_phone', $cardinality),
+      '#attributes' => [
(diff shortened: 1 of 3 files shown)


kalpaitch:
Let's have a chat about this. Two points, firstly that we got told at the first accessibility review not to use html5 'pattern' as it's not GDS compliant. Secondly as we've mentioned there are so many possible telephone formats that we need to be very loose on this...

jon-edery:
I was not aware of this "html5 'pattern' as it's not GDS compliant". we do use this in our date plugin... 

On the second point realistically the formats used for phone numbers on the register are standard ones.  We should have some basic validation on these fields.

kalpaitch:
"On the second point realistically the formats used for phone numbers on the register are standard ones." Are they? I don't think we know this for sure, we have done no research on this.

jon-edery:
I have reviewed contact data across the register I cannot see any of these potential edge cases where strings are needed to be used in combination with digits.  Do you have an example of such a case on the register?   seems unlikely 

I used the regex which is used by google to validate google input form:  ". Phone Numbers - these are often a series of numbers preceded by an optional ”+” sign and the area code may be inside brackets."

It might be better to have basic validation which covers 99% of cases then no validation.

PR #643, opened 2020-07-28
PAR-1637: Made email a required field for all contacts

kalpaitch:
I've made the email a required field for all contacts.

@jon-edery this also will now throw an error if we try to save an entity which doesn't have all the required properties. On the whole I think this is probably a good thing, but there may be some situations where we are creating empty entities for updating later.

My thoughts on this are that if we are then that field probably shouldn't be required in the first place, and the tests should show us where this is happening. But this is a bit of a heavy handed approach. Thoughts?

### DIFF
--- a/web/modules/custom/par_data/par_data.install
+++ b/web/modules/custom/par_data/par_data.install
@@ -1044,3 +1044,17 @@ function par_data_update_8044() {
     }
   }
 }
+
+/**
+ * PAR-1637 Update email field as required.
+ */
+function par_data_update_8045() {
+  $entity_definition_update_manager = \Drupal::entityDefinitionUpdateManager();
+
+  // New fields definitions to be updated to the par_data_person entity.
+  // @see https://www.drupal.org/node/3034742
+  $field_storage_definition = $entity_definition_update_manager->getFieldStorageDefinition('email', 'par_data_person');
+  $field_storage_definition->setRequired(TRUE);
+
+  $entity_definition_update_manager->updateFieldStorageDefinition($field_storage_definition);
+}
--- a/web/modules/custom/par_data/par_data.install
+++ b/web/modules/custom/par_data/par_data.install
@@ -1053,8 +1053,14 @@ function par_data_update_8045() {
 
   // New fields definitions to be updated to the par_data_person entity.
   // @see https://www.drupal.org/node/3034742
-  $field_storage_definition = $entity_definition_update_manager->getFieldStorageDefinition('email', 'par_data_person');
-  $field_storage_definition->setRequired(TRUE);
+  $person_field_storage_definition = $entity_definition_update_manager->getFieldStorageDefinition('email', 'par_data_person');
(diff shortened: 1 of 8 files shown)


kalpaitch:
This should resolve the issue of us being able to create multiple empty contact records. But on a second note we'll need to capture fatal error messages better so that we display a more friendly page to the users.

web/modules/custom/par_data/src/Entity/ParDataPerson.php
```diff
@@ -787,7 +787,7 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
     $fields['email'] = BaseFieldDefinition::create('string')
       ->setLabel(t('E-mail'))
       ->setDescription(t('The e-mail address of this person.'))
-      ->addConstraint('par_required')
```
jon-edery, line 790: Do we not need to have par_required as an added constraint for the GDS compliant form validation in addition to the set required?

jon-edery:
I think the issue you highlighted is the main concern here, also should we be identifying the point in the code were duplicate users are being created so this can be prevented at the source?

web/modules/custom/par_data/src/Entity/ParDataPerson.php
kalpaitch: Nah, one or other.

kalpaitch:
> I think the issue you highlighted is the main concern here, also should we be identifying the point in the code were duplicate users are being created so this can be prevented at the source?

In the case of using the back button the only other place we can change this is in the Review Forms and/or Form Plugins, and there are lots of them.

I'm tempted to think that would _also_ be a good place to make changes and to catch attempts to create empty data. I think it would be really good to have a universal way to identify whether a form has data, I guess this is partly where the idea of 'starting' a journey comes from, because then we can initiate a 'journey started' property. For example, do we have data for all the required steps in a journey, and do we have data for the current step?

web/modules/custom/par_data/par_data.install
```diff
@@ -1044,3 +1044,23 @@ function par_data_update_8044() {
     }
   }
 }
+
+/**
+ * PAR-1637 Update email field as required.
```
jon-edery, line 1049: person email and inspection_plan summary fields...

web/modules/custom/par_data/par_data.install
```diff
@@ -1044,3 +1044,23 @@ function par_data_update_8044() {
...
+ */
+function par_data_update_8045() {
+  $entity_definition_update_manager = \Drupal::entityDefinitionUpdateManager();
+
+  // New fields definitions to be updated to the par_data_person entity.
+  // @see https://www.drupal.org/node/3034742
+  $person_field_storage_definition = $entity_definition_update_manager->getFieldStorageDefinition('email', 'par_data_person');
+  $person_field_storage_definition->setRequired(TRUE);
+
+  // New fields definitions to be updated to the par_data_person entity.
```
jon-edery, line 1059: remove comment here.

PR #644, opened 2020-07-29
Error handling message.

kalpaitch:
Thought it was about time we had a better error screen than the epic WSOD:
Screenshot from 2020-07-29 10-58-47

So now we have (we should probably get the wording checked):
Screenshot from 2020-07-29 10-52-58

This also generates a unique code that helps us identify the error more clearly, this bit is obviously optional we don't need this:
Screenshot from 2020-07-29 10-56-11


kalpaitch:
Let's hold off this one a bit, get some approval from OSU first.

ralph-hawkins:
The GOV.UK Design System has a [pattern for 'problem with the service pages'](https://design-system.service.gov.uk/patterns/problem-with-the-service-pages/). The page should follow that layout as closely as possible.

ralph-hawkins:
Is the error code generated each time there is error?

kalpaitch:
> Is the error code generated each time there is an error?

Yea, I don't particularly want to show it but put it in there to indicate that we'll be providing the heldesk with a lit of recent errors.

ralph-hawkins:
Have you spoken to anyone about whether that will be a useful thing to help identify what happened?

kalpaitch:
> Have you spoken to anyone about whether that will be a useful thing to help identify what happened?

We have been unable to identify issues in the past because the user or the OSU member has been unable to provide us with sufficient information about when the error occurred. I think it is sufficient that we're now logging this information, I don't think we actually need to show the number to id to the end user though.

ralph-hawkins:
In that case, I think its probably best to go with the content as it is on the design system.

kalpaitch:
Screenshot from 2021-01-04 10-11-56

PR #648, opened 2020-08-13
PAR-1673 reg functions

jon-edery:
Prevent users creating authorities without regulatory functions.  
Screen Shot 2020-08-13 at 12 53 26


kalpaitch (approved):
Seems good to me, let's get this tested in the next release.

PR #649, opened 2020-08-14
PAR-1671  Advice title

jon-edery:
Use the advice title if one has been set.  In case the advice does not have a title set i.e. old advice documents fallback to the attachment filename.  

Screen Shot 2020-08-18 at 11 20 41

Screen Shot 2020-08-18 at 11 20 01


kalpaitch (approved):
Works for me!

PR #651, opened 2020-08-14
PAR-1670 archive reason validation

jon-edery:
Screen Shot 2020-08-13 at 18 51 09

### DIFF
--- a/web/modules/custom/par_data/par_data.install
+++ b/web/modules/custom/par_data/par_data.install
@@ -1044,3 +1044,10 @@ function par_data_update_8044() {
     }
   }
 }
+
+/**
+ * PAR-1617 refresh archive reason field.
+ */
+function par_data_update_8046() {
+  \Drupal::service('entity.definition_update_manager')->applyUpdates();
+}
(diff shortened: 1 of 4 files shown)


web/modules/custom/par_data/par_data.install
```diff
@@ -1044,3 +1044,10 @@ function par_data_update_8044() {
     }
   }
 }
+
+/**
+ * PAR-1610 refresh archive reason field.
+ */
+function par_data_update_8046() {
+  \Drupal::service('entity.definition_update_manager')->applyUpdates();
```
kalpaitch, line 1052: // New field definitions to be added to the system entities.
// @see https://www.drupal.org/node/3034742

I'm not sure the catchall `applyUpdates()` method is working any more.

jon-edery: fair point.

jon-edery: It does work (have tested) its not recommended however lets do it the long annoying tho.

PR #653, opened 2020-08-19
PAR-1545: Accessibility issue WCAG2.4.2 Page titled

kalpaitch:
This resolves issues with the dashboard page and the homepage sharing a name = 'Primary Authority Register'

We have changed the dashboard title to Dashboard. And also introduced some logic to make sure the site name 'Primary Authority Register' is a) de-prioritised in the html title tag b) no repeated in the html title tag.

### DIFF
--- a/web/themes/custom/par_theme/sass/par.scss
+++ b/web/themes/custom/par_theme/sass/par.scss
@@ -287,6 +287,7 @@ a.flow-link {
 fieldset, legend {
   margin: 0px;
   padding: 0px;
+  max-width: 100%;
 }
 
 .heading-medium.fieldset-legend {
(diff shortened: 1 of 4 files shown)


kalpaitch:
Woops, I also combined PAR-1543 into this PR. I guess it's not too important though.

web/themes/custom/par_theme/sass/par.scss
```diff
@@ -287,6 +287,7 @@ a.flow-link {
 fieldset, legend {
   margin: 0px;
   padding: 0px;
+  max-width: 100%;
 }
```
jon-edery, line 291: You may need to increase the version number in the CI config to force the CSS to re-compile on deployment.

kalpaitch: True, but it's not failing tests anyway so who cares what CI says :)

PR #663, opened 2020-09-11
PAR-1680: Theme upgrade spike.

kalpaitch:


kalpaitch:
@nasirkhan1982 I've added a new base theme based on the latest govuk styles. This can be enabled by adding `$config['config_split.config_split.theme_v2_config']['status'] = TRUE;` to your local settings file.

The good news is this base theme works in principle:
Screenshot from 2021-01-04 12-40-11

We'll need to plan the rest of the upgrade carefully so that we catch all the remaining elements and components that need to be converted. I thought I'd started a list of these somewhere, I'll take another look.

PR #679, opened 2020-11-24
PAR-1686: Initial work to migrate inspection plan and advice files to media

kalpaitch:
The first tranche of work consists of migrating just the advice and inspection plan files to media entities. Provided this works as expected the remaining PAR entities can be migrated.

@TODO Rebuild affected test data.


kalpaitch:
Have separated the migrations scripts and config from this PR, such that the scripts can be deployed before the code is switched to use the new files.

PR #685, opened 2020-12-10
PAR-1691: Upgraded Default Content Module

kalpaitch:
Upgraded default content module to version 2.0, primarily because:
* It supports files without file_entity (need to check this)
* It has a more compact output, which makes it easier to create new test data


nasirkhanbeis (approved):
looks good!

PR #841, opened 2022-08-08
PAR-1783 Change ParDataPartnership::addLegalEntity() to create new pa…

RogerBailey:
…rtnership_legal_entity linking the partnership with the newly created legal_entity.

Tested the adding of a new legal entity to the partnership. Inspection of the database confirmed that all records created correctly.

### DIFF
--- a/web/modules/custom/par_data/src/Entity/ParDataPartnership.php
+++ b/web/modules/custom/par_data/src/Entity/ParDataPartnership.php
@@ -683,13 +683,8 @@ class ParDataPartnership extends ParDataEntity {
     $partnership_legal_entity->setStartDate(new DrupalDateTime());
     $partnership_legal_entity->save();
 
-    // Retrieve existing partnership legal entities.
-    $partnership_legal_entities = $this->getPartnershipLegalEntity();
-
-    // Append new partnership legal entity to existing entities.
-    $partnership_legal_entities[] = $partnership_legal_entity;
-
-    $this->set('field_partnership_legal_entity', $partnership_legal_entities);
+    // Append new partnership legal entity to existing list of legal entities covered by this partnership.
+    $this->get('field_partnership_legal_entity')->appendItem($partnership_legal_entity);
   }
 
   /**
--- a/web/modules/custom/par_data/src/Entity/ParDataPartnership.php
+++ b/web/modules/custom/par_data/src/Entity/ParDataPartnership.php
@@ -677,10 +677,16 @@ class ParDataPartnership extends ParDataEntity {
    */
   public function addLegalEntity(ParDataLegalEntity $legal_entity) {
 
-    // Create new partnership legal entity.
+    // Create new partnership legal entity referencing the legal entity.
     $partnership_legal_entity = ParDataPartnershipLegalEntity::create([]);
     $partnership_legal_entity->addLegalEntity($legal_entity);
-    $partnership_legal_entity->setStartDate(new DrupalDateTime());
+
+    // Start date only needed if partnership is already approved.
+    // If start date not set then it is assumed that the legal entity's
+    // participation starts when the partnership starts.
+    if ($this->isActive()) {
+      $partnership_legal_entity->setStartDate(new DrupalDateTime());
+    }
     $partnership_legal_entity->save();
 
     // Append new partnership legal entity to existing list of legal entities covered by this partnership.
--- a/web/modules/custom/par_data/src/Entity/ParDataPartnership.php
+++ b/web/modules/custom/par_data/src/Entity/ParDataPartnership.php
@@ -660,19 +660,36 @@ class ParDataPartnership extends ParDataEntity {
   }
 
   /**
+   * Get partnership legal entities for this partnership.
+   */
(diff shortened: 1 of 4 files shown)


web/modules/custom/par_data/src/Entity/ParDataPartnership.php
```diff
@@ -659,20 +659,37 @@ public function getLegalEntity($single = FALSE) {
...
+    $partnership_legal_entity->save();
+
+    // Retrieve existing partnership legal entities.
+    $partnership_legal_entities = $this->getPartnershipLegalEntity();
+
+    // Append new partnership legal entity to existing entities.
+    $partnership_legal_entities[] = $partnership_legal_entity;
 
-    $this->set('field_legal_entity', $legal_entities);
+    $this->set('field_partnership_legal_entity', $partnership_legal_entities);
```
kalpaitch, line 692: One liner that can replace the last 3 lines here `$this->get('field_partnership_legal_entity')->appendItem($partnership_legal_entity);`

web/modules/custom/par_data/src/Entity/ParDataPartnership.php
```diff
@@ -659,20 +659,37 @@ public function getLegalEntity($single = FALSE) {
...
    *   A PAR Legal Entity to add.
    */
   public function addLegalEntity(ParDataLegalEntity $legal_entity) {
-    // Retrieve existing legal entities.
-    $legal_entities = $this->getLegalEntity();
 
-    // Append new legal entity to existing entities.
-    $legal_entities[] = $legal_entity;
+    // Create new partnership legal entity.
+    $partnership_legal_entity = ParDataPartnershipLegalEntity::create([]);
```
kalpaitch, line 681: Just thinking aloud here, do we need to add any validation rules here to make sure legal entities cannot be added more than once to a partnership?

web/modules/custom/par_data/src/Entity/ParDataPartnership.php
```diff
@@ -659,20 +659,37 @@ public function getLegalEntity($single = FALSE) {
...
+    $partnership_legal_entity->addLegalEntity($legal_entity);
+    $partnership_legal_entity->setStartDate(new DrupalDateTime());
```
kalpaitch, line 683: Also, another thought here on validation, do we need to check whether the partnership is approved yet, because technically the start date is only really need if we're adding the legal entity after the partnership has been approved.

If we're adding a legal entity before the partnership is approved then we don't really need to set a date (we can assume that no date means the legal entity starts when the partnership starts).

web/sites/default/settings.php
```diff
@@ -1128,3 +1128,15 @@
   ini_set('memory_limit', '4G');
   ini_set('max_execution_time', '3600');
 }
+$databases['default']['default'] = array (
```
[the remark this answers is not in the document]
RogerBailey: Oh yes. Will remove.

web/modules/custom/par_data/src/Entity/ParDataPartnership.php
RogerBailey: Yes. Will change this. Had just copied the code from elsewhere.

web/modules/custom/par_data/src/Entity/ParDataPartnership.php
RogerBailey: I think it is OK to have a legal entity appearing more than once as long as the periods defined by start and end dates do not overlap.

Would the correct way to add validation to be to override the Trance::preSave() method and throw a ParDataException if an inconsistency is found.

web/modules/custom/par_data/src/Entity/ParDataPartnership.php
RogerBailey: That does make sense. No need to setStartDate() if partnership is not yet approved. Will make that change.

web/modules/custom/par_data/src/Entity/ParDataPartnership.php
```diff
@@ -659,20 +659,38 @@ public function getLegalEntity($single = FALSE) {
...
-    // Append new legal entity to existing entities.
-    $legal_entities[] = $legal_entity;
+    // Create new partnership legal entity referencing the legal entity.
+    $partnership_legal_entity = ParDataPartnershipLegalEntity::create([]);
+    $partnership_legal_entity->addLegalEntity($legal_entity);
+
+    // Start date only needed if partnership is already approved.
+    // If start date not set then it is assumed that the legal entity's
+    // participation starts when the partnership starts.
+    if ($this->isActive()) {
```
kalpaitch, line 687: Soz, I think isApproved() is the right method.

kalpaitch:
Looks great!

web/modules/custom/par_data/src/Entity/ParDataPartnership.php
RogerBailey: Made the change using partnership->isActive() to check whether start date should be set. Hard to test properly since route partnership/xxx/legal_entity/add is currently not valid when partnership is active.

PR #843, opened 2022-08-11
Par 1781 migrate relationship data to new meta data entity

RogerBailey:
Changes made per your and mine comment on earlier commits.


web/modules/custom/par_data/src/Entity/ParDataPartnership.php
```diff
@@ -679,7 +679,7 @@ public function addLegalEntity(ParDataLegalEntity $legal_entity) {
 
     // Create new partnership legal entity referencing the legal entity.
     $partnership_legal_entity = ParDataPartnershipLegalEntity::create([]);
-    $partnership_legal_entity->addLegalEntity($legal_entity);
+    $partnership_legal_entity->setLegalEntity($legal_entity);
```
kalpaitch, line 682: I think this shouldn’t change, its probably used elsewhere in code in suspect, and either way this should add multiple legal entities anyway.

kalpaitch (commented):
I think we should release what we have already in PAR-788 first. And then your update hook should work.

web/modules/custom/par_data/src/Entity/ParDataPartnership.php
RogerBailey: I think you are confusing ParDataPartnership::addLegalEntity() with ParDataPartnershipLegalEntity::addLegalEntity(). For the former each call to addLegalEntity() will create a new partnership_legal_entity instance and add it to the list of references. For the later the method is setting the reference to the ParDataLegalEntity instance on the ParDataPartnershipLegalEntity instance and discarding any previous reference. This is why I think it should be named setLegalEntity() not addLegalEntity().

kalpaitch: My bad, I totally thought this function name had changed. I’ll read more carefully next time :)

PR #854, opened 2022-08-25
PAR-1773: Updated the search index and the OSU advanced partnership search view

kalpaitch:
Also includes PAR-1799 and half of PAR-1768.

In par_data_views_data_alter have converted the PAR fields that have allowed_values to use the search_api_options filter and have adapted the options callback to lookup the PAR allowed_values from our entity config.

Have also adapted the entity reference fields that have less than 20 items (e.g. regulatory functions) to use the search_api_options too.


RogerBailey:
Had a look through the changes. All looks consistent and nothing stands out as an issue.

PR #865, opened 2022-09-30
PAR-788: Partnership change requests

kalpaitch:
Adds the ability to modify the legal entities on an existing partnership.


RogerBailey (changes requested):
I thin line 1418 should be `$partnership->set('member_link', 'N');`

RogerBailey (approved):
All good. Didn't understand that NULL get saved as 'N' in the subfield.

PR #867, opened 2022-10-10
Par 1906 user able to add duplicate legal entity after reinstating

RogerBailey:
Partnership legal entity reinstate form can no longer be accessed if the PLE can not be reinstated.

Add validation to partnership legal entity reinstate form to double check that reinstating is OK.


kalpaitch (approved):
Noice

PR #904, opened 2023-01-10
complete add enforcement journey

ndileonel:


tests/e2e/par-test-automation/link.txt
```diff
@@ -0,0 +1 @@
+https://www.udemy.com/course/foundation-level-training/learn/lecture/22297150#reviews
```
kalpaitch, line 1: If this is a test link could we make it example.com or something?

tests/e2e/par-test-automation/src/test/resources/features/PAR-General.feature
```diff
@@ -3,7 +3,7 @@ Feature:
   I  want to be able to view/manage partnerships
   So I can comply with the BEIS standards for goods and services
   
- @regression @partnershipapplication @direct @update @usermanagement @organisation
+ @regression @partnershipapplication @direct @update @usermanagement @organisation @enforcement
```
kalpaitch, line 6: Could we discuss tagging all features with ci and smoke tags or something like this so that it's obvious which ones to run...

tests/e2e/par-test-automation/src/test/resources/features/PAR-General.feature
```diff
@@ -90,13 +89,25 @@ Feature:
...
   And the user logs in with the "par_helpdesk@example.com" user credentials
   When the user searches for the last created organisation
   And the user updates all the fields for last created organisation
   Then all the fields are updated correctly
   
+ @regression @enforcement
+ Scenario: Verify Send Notification of Proposed Enforcement (Happy Path - PAR-1852)
+  Given the user is on the PAR login page
+  And the user logs in with the "par_enforcement_officer@example.com" user credentials
+  When the user searches for the last created partnership
```
kalpaitch, line 104: I'm constantly thinking about set up data, and being able to run tests in any order, lord forgive me!

If we were to run just this scenario, then there may not be a 'last created partnership'. Could we maybe discuss how we can best handle the setup of for tests like this.

ndileonel: sure thing - i'm guessing you're referring to using magic data. we could write separate test scenarios just to run specific functionality using setup data. we'll have to decide which tests we want to use magic data
Note: currently, it'll cost ONLY creating a partnership and nominating it (~2 mins), to be able to run this test, and pretty much for the bulk of the other tests requiring a nominated partnership as long as we're tagging the tests efficiently

tests/e2e/par-test-automation/src/test/resources/features/PAR-General.feature
ndileonel: agree

tests/e2e/par-test-automation/link.txt
ndileonel: changed accordingly

tests/e2e/par-test-automation/src/test/resources/features/PAR-General.feature
ndileonel: had discussion with Andrew already on this. will update tests with tags accordingly when we commence work with circleci integration

PR #930, opened 2023-03-07
Par 2036

ndileonel:


RogerBailey (changes requested):
Files InspectionFeedbackSearchPage.java and PAR-General.feature are being flagged as not having a new line character at the end of the file.

Otherwise looks OK to me.

PR #974, opened 2023-06-19
Accessibility remedial

kalpaitch:


composer.json
```diff
@@ -25,14 +25,16 @@
         }
     ],
     "require": {
-        "php": "~8.1.0",
+        "php": "~8.1.2",
```
kalpaitch, line 28: @robwildolivejar It was actually this change to the compose.json file.

composer.json
```diff
@@ -32,7 +32,9 @@
         "composer/installers": "^1.9",
         "cweagans/composer-patches": "^1.7",
         "deblan/csv-validator": "dev-master",
+        "drupal/a11y_autocomplete_element": "^1.0",
         "drupal/address": "~1.11.0",
+        "drupal/admin_toolbar": "^3.4",
```
kalpaitch, line 37: Not sure this is strictly necessary, we don't really have admin users (aside from us devs)??

composer.json
```diff
@@ -179,7 +181,9 @@
                 "Form API #ajax not working on link element": "https://www.drupal.org/files/issues/2023-01-11/core-allow-form-ajax-for-links-2915954-9.5.1.patch",
                 "Add events for matching entity hooks": "https://www.drupal.org/files/issues/2022-12-22/2551893-99.patch",
                 "Non-exposed and exposed default sort order": "https://www.drupal.org/files/issues/2021-02-16/3120847-20.patch",
-                "PostgreSQL: email case insensitivity": "https://www.drupal.org/files/issues/2022-12-19/drupal-case_sensitive_email-2490294-58-d95.patch"
+                "PostgreSQL: email case insensitivity": "https://www.drupal.org/files/issues/2022-12-19/drupal-case_sensitive_email-2490294-58-d95.patch",
+                "Add aria labelling for form states": "./patches/states-aria.patch",
```
kalpaitch, line 185: Let's try and create a drupal issue for each of these patches and link to this. That way we will have a record of why the patch has been applied and what it does.

package.json
```diff
@@ -22,7 +21,8 @@
   },
   "scripts": {
     "gulp": "gulp",
-    "install-govuk-theme": "cd web/themes/contrib/govuk_theme && npm install && npm run gulp build -- --ce --ct --cf --cl",
-    "install-par-theme": "cd web/themes/custom/govuk_par && npm install"
+    "install-govuk-par": "cd web/themes/custom/govuk_par && npm install && npm run gulp",
```
kalpaitch, line 24: Will review this module separately to see why contrib module doesn't meet the requirements here. Could always try and add you as a maintainer of this...

patches/inline_form_errors.patch
```diff
@@ -0,0 +1,218 @@
+diff --git a/core/lib/Drupal/Core/Form/FormElementHelper.php b/core/lib/Drupal/Core/Form/FormElementHelper.php
```
kalpaitch, line 1: I feel like I've seen a drupal.org issue for this, could you share if there is anything this is based on.

sync/block.block.analyticscookies_2.yml
```diff
@@ -0,0 +1,25 @@
+uuid: dd547e01-7e4c-40fb-ae49-3f00110d02f6
```
kalpaitch, line 1: Why is there a second analytics cookies block?

I guess one for each of the themes, just trying to work out why this has changed when the contrib govuk_theme didn't have one. I'm not sure how these two themes link together at this point, will have a bit more of a dig, any light you can shed on this?

web/modules/custom/par_flows/src/ParRedirectTrait.php
```diff
@@ -75,7 +75,7 @@ public function getLinkByRoute($route, $route_params = [], $link_options = [], $
   public function getLinkByUrl(Url $url, $text = '', $link_options = []) {
     $link_options += [
       'absolute' => TRUE,
-      'attributes' => ['class' => 'flow-link']
+      'attributes' => ['class' => 'govuk-link']
```
kalpaitch, line 78: Completely agree with adding the class, but lets keep the other class there for now, no need to introduce more breaking changes than necessary.

web/modules/custom/par_forms/src/Plugin/ParForm/ParBusinessSizeForm.php
```diff
@@ -42,7 +42,7 @@ public function getElements($form = [], $cardinality = 1) {
 
     $form['info'] = [
       '#markup' => t('Enter the number of associations in your membership list'),
```
kalpaitch, line 44: If we're making this change we could probably replace it with the more appropriate render element:
```
          '#type' => 'html_tag',
          '#tag' => 'h2',
          '#attributes' => ['class' => ['govuk-heading-m']],
          '#value' => $this->t('Enter the number of associations in your membership list'),
```

web/modules/custom/par_forms/src/Plugin/ParForm/ParContactDisplay.php
```diff
@@ -98,12 +98,12 @@ public function loadData($cardinality = 1) {
...
       $form['message_intro'] = [
-        '#type' => 'fieldset',
-        'title' => [
+        '#type' => 'container',
+        'heading' => [
           '#type' => 'html_tag',
           '#tag' => 'h2',
           '#value' => $this->t('Contacts'),
-          '#attributes' => ['class' => ['heading-large']],
+          '#attributes' => ['class' => ['z']],
```
kalpaitch, line 106: Little oopsie here, think the class should be `govuk-heading-l`

web/modules/custom/par_forms/src/Plugin/ParForm/ParContactDisplay.php
```diff
@@ -143,7 +143,7 @@ public function getElements($form = [], $cardinality = 1) {
       }
 
       $form['contact'] = [
-        '#type' => 'fieldset',
+        '#type' => 'container',
         '#weight' => 1,
         '#attributes' => ['class' => ['grid-row', 'form-group', 'contact-details']],
```
kalpaitch, line 148: I can see a `grid-row` class here that should probably be updated to `govuk-grid-row` as per the other changes. Maybe just one final check that all these classes have been changed??

web/modules/custom/par_forms/src/Plugin/ParForm/ParGdprForm.php
```diff
@@ -64,6 +64,7 @@ public function getElements($form = [], $cardinality = 1) {
 
     $form['summary'] = [
       '#theme' => 'item_list',
+      '#list_header_tag' => 'h2',
```
kalpaitch, line 67: Looks great. I was wondering if this applies to any other elements (particularly those where there is just one question on a page and the label should become the primary heading) ?? Probably a separate conversation though.

web/modules/custom/par_notification/src/Controller/ParAccessDeniedController.php
```diff
@@ -63,11 +63,11 @@ public function build(Request $request) {
       && !user_load_by_mail($message->get('field_to')->getString()));
 
     // Column size is determined by whether invitations should be shown.
-    $column_size = $allow_invitations ? 'column-one-half' : 'column-full';
+    $column_size = $allow_invitations ? 'column-one-half' : 'govuk-grid-column-full';
```
kalpaitch, line 66: Need to change the `column-one-half` class as well...

web/modules/features/par_enforcement_raise_flows/src/Form/ParChecklistForm.php
```diff
@@ -89,7 +89,7 @@ public function buildForm(array $form, FormStateInterface $form_state) {
         '#title' => $this->t('Discuss this enforcement'),
         '#url' => Url::fromRoute('par_enquiry_send_flows.select_authority', $this->getRouteParams()),
         '#prefix' => '<p class="form-group">',
-        '#suffix' => '<br><br><br></p>',
+        '#suffix' => '</p>',
```
kalpaitch, line 92: Just a reminder to check spacing on this element.

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsSicCodeForm.php
```diff
@@ -59,7 +59,7 @@ public function buildForm(array $form, FormStateInterface $form_state, ParDataPa
     }
 
     $form['sic_code'] = [
-      '#type' => 'select',
+      '#type' => 'a11y_autocomplete',
```
kalpaitch, line 62: I see this element change is causing a bit of trouble, will review this separately as well. There is a lot of work to do on the sic code selection, and I don't think this element is necessarily the solution going forward, may be better just to create a separate ticket for this work??

web/modules/custom/par_notification/src/Controller/ParAccessDeniedController.php
kalpaitch: Could be useful to create a linting plugin/script to check we've replaced all the old class names (if it's a simple task).

composer.json
robwildolivejar: Not strictly necessary but super useful - I enable it for every task I pick up

composer.json
robwildolivejar: The postgreSQL is already there, the aria-labels one I will do

package.json
robwildolivejar: Its been a long time since any work was done for this

sync/block.block.analyticscookies_2.yml
robwildolivejar: One for each enabled theme

web/modules/custom/par_flows/src/ParRedirectTrait.php
robwildolivejar: No need to keep legacy classes, the testers have updated the tests accordingly

web/modules/custom/par_forms/src/Plugin/ParForm/ParContactDisplay.php
robwildolivejar: Good spot, have fixed this

web/modules/custom/par_forms/src/Plugin/ParForm/ParContactDisplay.php
robwildolivejar: Good spot, I have changed this and searched the rpo for other occurrences

web/modules/features/par_enforcement_raise_flows/src/Form/ParChecklistForm.php
robwildolivejar: The artificial padding this adds is now picked up via CSS and applied correctly, albeit on the form-group elements in a one of the 4 types of row styles

web/modules/features/par_partnership_flows/src/Form/ParPartnershipFlowsSicCodeForm.php
robwildolivejar: the SIC code is now back as a select

robwildolivejar (commented):
Review supdated

robwildolivejar (commented):
Updated merge regression

PR #1036, opened 2023-09-22
PAR-2283

ThommCroft:
Added automation test for PAR-1868 to verify the addition of a Member Organisation to a coordinated partnership.
Added the coordinated user and Business user credentials to the par-config file.
All tests pass as expected in a full Regression test.


kalpaitch:
This does need the conflict resolved.

